Issue #321, opened 2015-01-18
If Scope extends `MustExpectations` - it causes failing tests to look like running and **passing**


maximn:
Here's a sample code -
``` scala
package com.sample

import org.specs2.matcher.MustExpectations
import org.specs2.mutable.SpecificationWithJUnit
import org.specs2.specification.Scope


class BadTest extends SpecificationWithJUnit {

  trait ctx extends Scope with MustExpectations {}

  "Wrong statement" should {
    "be failed by a test" in new ctx {
      true must be_===(false)
    }
  }
}


//trait ctx extends Scope with Matchers with MustExpectations {
//  def someThingThatUsesMust(x: AnyRef) = x must beNull
//}

```

As you can see in the screenshot - IntelliJ shows the test as running and green.
![IntelliJ screenshot](https://www.evernote.com/shard/s203/sh/5724bdd8-5fd1-4ff6-a149-72f923e0b406/4162cb9400269c21/res/812c08f3-ce08-49b4-9ed6-833d157c2d5c/BadTest_scala_-__testapp__-_TestApp_-____Code_TestApp_.png?resizeSmall&width=832)

etorreborre:
Yes, it needs to extend `MustThrownExpectations` for exceptions to be thrown and caught outside of the `Scope` trait.

etorreborre:
Also, in this specific example you shouldn't need to create `MustExpectations` because `MustThrownExpectations` is already mixed in `org.specs2.mutable.SpecificationWithJUnit` and allows to create expectations already. So I guess that you are creating the `ctx` trait outside of a specific `Specification` to be able to reuse it, right?

maximn:
Maybe I shouldn't use `MustExpectations`, but if I do, I expect to **get an error** somewhere.
Yes, I use it outside. Actually in my case my context extends a trait that extends `MustExpectations` but I wanted to provide the most simple example.

etorreborre:
There are 2 traits offering the same API but having different implementations: `MustExpectations` and `MustThrownExpectations`. As the name indicates `MustThrownExpectations` must be used where exceptions are expected for control flow. Typically in this situation where code is being placed in the body of a trait (the `ctx` trait); if no exception is thrown there is no way to get notified of failures.

It would be better to have a compilation error when `MustExpectations` is being mixed-in with `Scope` but I don't know how to do that.

etorreborre:
[...]

PR #324, opened 2015-01-21
Don't allow mixin Scope with MustExpectations

maximn:
https://github.com/etorreborre/specs2/issues/321

### DIFF
--- a/matcher/src/main/scala/org/specs2/matcher/ThrownExpectations.scala
+++ b/matcher/src/main/scala/org/specs2/matcher/ThrownExpectations.scala
@@ -5,7 +5,6 @@ import execute._
 import execute.Skipped
 import execute.Pending
 import execute.Failure
-import scala.Some
 import execute.PendingException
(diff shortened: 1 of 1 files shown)


etorreborre:
I don't see how to do a reasonable compile-time check so I added your runtime check for now (slightly generalised to `ShouldMatchers`). Thanks.

Nimrod007:
Hi, in the end this was not merged ? 
because i see this in the sourcecode:   
https://github.com/etorreborre/specs2/blob/226d84bca61b3dca093af87c5846e042e43ecb0b/matcher/src/main/scala/org/specs2/matcher/ThrownExpectations.scala#L127-L130

etorreborre:
This was not merged directly because I changed the code slightly. Instead of just checking for `MustExpectations` and `MustThrownExpectations`, I check `Expectations` and `ThrownExpectations`. This allows the `ShouldExpectation` trait to be supported as well.

PR #339, opened 2015-03-07
Feature/escape xml

alissapajer:
Escapes the four characters: < > & "


project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.1")
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
```
alissapajer: According to [the docs](http://www.scala-sbt.org/sbt-pgp/), this version of `sbt-pgp` can be used for sbt versions 0.13.5+

etorreborre:
Thanks Alissa, it was a terrible oversight for me to not escape properly the strings.

PR #340, opened 2015-03-07
Escape all the things! Only then shall the xml be valid.

alissapajer:
The bodies of all the xml fields must also be escaped.


etorreborre:
Escape all the things indeed :+1:

alissapajer:
Could you publish a release with this fix in it? Then I'll try it out! :)

etorreborre:
I did an [official release](http://notes.implicit.ly/post/113113471389/specs2-3-0-1) yesterday. I hope it will work fine for you.

alissapajer:
Ah great, trying that now!

PR #375, opened 2015-05-13
make test-only command camelCased

taisukeoe:
Since some parts of document keep `test-only` command hyphenated, this PR makes them camelCased. 

Ref:
http://www.scala-sbt.org/0.13.0/docs/Community/ChangeSummary_0.13.0.html#camelcase-key-names


etorreborre:
Thanks!

PR #388, opened 2015-06-25
Introduced responds method that allows to use anonymous case function for a mock

dkomanov:
Cannot use anonymous case function for answers because of overloading (it requires signature specified, so it's necessary to write more code: answers { _ match { case i: Int => (i + 1).toString } }


etorreborre:
Can you please add a note in the [User Guide](https://github.com/etorreborre/specs2/blob/master/guide/src/test/scala/org/specs2/guide/UseMockito.scala#L119) as well?

dkomanov:
Done.

etorreborre:
Thanks!

PR #409, opened 2015-08-31
Update scalaz dependency to 7.1.3

kflorence:
Addresses https://github.com/scalaz/scalaz/pull/932

Took me a couple of hours to figure out why I was getting `java.lang.reflect.MalformedParameterizedTypeException` exceptions when trying to autowire beans in my test classes. Traced it down to the above pull request which fixes the issue in Scala 2.11.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,19 +11,19 @@
         <dependency>
             <groupId>org.scalaz</groupId>
             <artifactId>scalaz-core_2.10</artifactId>
-            <version>7.1.1</version>
+            <version>7.1.3</version>
             <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.scalaz</groupId>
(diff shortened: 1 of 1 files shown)


etorreborre:
Do you need a snapshot release?

kflorence:
@etorreborre Thanks, that would be great if you don't mind.

etorreborre:
Done as `3.6.4-20150831233848-1f41c5e`.

kflorence:
@etorreborre Much appreciated :+1:

kflorence:
@etorreborre Can you publish a snapshot for 2.10.x as well when you get the chance? We are cross-compiling our libraries so it would be helpful.

etorreborre:
Sorry for the delay, here it is: `3.6.4-20150901013911-1f41c5e`.

PR #417, opened 2015-09-26
Only fail on a future if Await.result timeouts

knutwalker:
This seems like an unnecessary check, but if the Future fails with some `T <: TimeoutException`, this exception would be swallowed and wrongly reported as a Failure.
I got this recently when trying to test some akka ask behavior and they fail with a [AskTimoeutException](https://github.com/akka/akka/blob/v2.3.14/akka-actor/src/main/scala/akka/pattern/AskSupport.scala#L21). Using `await`, I couldn't test that this particular Exception was failing the Future.
The proposed fix only works for actual subclasses, not if some other code fails with a `TimeoutExeption` directly.

### DIFF
--- a/matcher/src/main/scala/org/specs2/matcher/FutureMatchers.scala
+++ b/matcher/src/main/scala/org/specs2/matcher/FutureMatchers.scala
@@ -48,10 +48,10 @@ trait FutureBaseMatchers extends ExpectationsCreation {
       def awaitFuture(remainingRetries: Int, totalDuration: FiniteDuration): Result = {
         try Await.result(f.map(value => AsResult(value))(ee.executionContext), appliedTimeout)
         catch {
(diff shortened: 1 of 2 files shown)


etorreborre:
Ok, I think I understand the purpose of the change. We only want to retry exceptions which are truly `TimeoutException`s and not any subclass. In that case I think you can simply write:
```
case e if e.getClass == classOf[TimeoutException] =>
```
Also it would be nice to add a test case in the `org.specs2.matcher.FutureMatchersSpec` specification to show what happens when there is a subclass of `TimeoutException` being thrown. Thanks.

knutwalker:
Thanks for the suggestions, I amended the commit accordingly.

etorreborre:
Great, thanks!

PR #427, opened 2015-10-26
Implements selection to run a test multiple times.

alissapajer:
For example:
`sbt test-only *MySpec* -- times 10`
will run `MySpec` ten times in a row!

Motivation for this: When testing nondeterministic code, sometimes a test will only sometimes fail. In that case, my current way to run a spec multiple times is:

`; test-only *MySpec*; test-only *MySpec*; test-only *MySpec*; ...`

And that's tedious.

### DIFF
--- a/core/src/main/scala/org/specs2/specification/process/Executor.scala
+++ b/core/src/main/scala/org/specs2/specification/process/Executor.scala
@@ -47,7 +47,13 @@ trait DefaultExecutor extends Executor {
    *  - sequence the execution so that only parts in between steps are executed concurrently
    */
   def execute(env: Env): Process[Task, Fragment] => Process[Task, Fragment] = { contents: Process[Task, Fragment] =>
-     execute1(env)(contents).andFinally(Task.delay(env.shutdown))
+    val times: Int = env.arguments.times
+    val fragments: Process[Task, Fragment] = execute1(env)(contents)
+
+    val res: Process[Task, Fragment] =
+      Stream.continually(fragments).take(times - 1).foldLeft(fragments)(_ ++ _)
+
+    res.andFinally(Task.delay(env.shutdown))
   }
 
   /**
--- a/common/src/main/scala/org/specs2/main/Arguments.scala
+++ b/common/src/main/scala/org/specs2/main/Arguments.scala
@@ -26,6 +26,7 @@ case class Arguments (
   commandLine:   CommandLine      = CommandLine()
  ) extends ShowArgs {
   def ex: String                      = select.ex
+  def times: Int                      = select.times
   def include: String                 = select.include
   def exclude: String                 = select.exclude
   def keep(tags: String*)             = select.keep(tags:_*)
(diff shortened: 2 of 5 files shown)


core/src/main/scala/org/specs2/specification/process/Executor.scala
```diff
@@ -47,7 +47,13 @@ trait DefaultExecutor extends Executor {
...
    */
   def execute(env: Env): Process[Task, Fragment] => Process[Task, Fragment] = { contents: Process[Task, Fragment] =>
-     execute1(env)(contents).andFinally(Task.delay(env.shutdown))
+    val times: Int = env.arguments.times
+    val fragments: Process[Task, Fragment] = execute1(env)(contents)
+
+    val res: Process[Task, Fragment] =
+      Stream.continually(fragments).take(times - 1).foldLeft(fragments)(_ ++ _)
+
+    res.andFinally(Task.delay(env.shutdown))
```
alissapajer: @etorreborre wanted to check with you that running the spec `n` times in this way (concating the `Process[Task, Fragment]`) using the same `env` the whole time would be okay. For example, would this play nicely with `Before` and `After`? I was just thinking about that case now...

etorreborre:
Hi Alissa,

I would suggest something different, extending what you have in the `AroundEach` trait to do things before and after each example. You can use a similar implementation to re-run each example. Something like that:
```
import org.specs2.execute.{Result, AsResult}
import org.specs2.specification.create.{ContextualFragmentFactory, FragmentsFactory}
import org.specs2.specification.{Around, Context}
import org.specs2.specification.core.Env
import scala.annotation.tailrec

trait RepeatEach extends FragmentsFactory {
  override protected def fragmentFactory = new ContextualFragmentFactory(super.fragmentFactory, aroundContext)

  protected def aroundContext: Env => Context = (env: Env) =>
    new Around {
      def around[R : AsResult](r: =>R) =
        repeat(env.arguments.commandLine.intOr("times", 1), r)
    }

  /**
   * repeat the evalution of a result n times
   * and stop if it is not successful
   */
  @tailrec
  final def repeat[R : AsResult](n: Int, r: =>R): Result =
    if (n <= 1) AsResult(r)
    else {
      val result = AsResult(r)
      if (result.isSuccess) repeat(n - 1, r)
      else result
    }
}
```
Would that work for you?

alissapajer:
@etorreborre yes, I think that would work. What's most important is that I can set the number of times the test is run in sbt with a `-- times` parameter, for example. 

In your example, we'd get the argument from `env.arguments.commandLine`, so I could do just that, if I understand it correctly.

etorreborre:
@alissapajer yes that's the idea, you can pass the argument from the command line to control the number of times you want examples to be executed.

etorreborre:
@alissapajer I'm just thinking that my proposal might not be so clear. I am not sure if it makes sense yet to include this trait in the distribution. I was proposing that you use that code and experiment with it. Maybe there are other features you want to add to it. The important part is that:

 - there is a way to control the (re-)execution of examples 
 - you can control it from the command line

You can apply this to: timeouts, Hadoop execution (local vs cluster),...

alissapajer:
@etorreborre ok cool, I'll play around with that trait, see what else might be interesting to add, and then maybe at some point we can include it in the distro!

etorreborre:
I am closing this PR now since there is a work-around.

PR #432, opened 2015-11-12
Add matchers for scalaz.Validation

puffnfresh:


etorreborre:
That's great! I should have done that a long time ago. Here is a pre-release `3.6.5-20151112214348-18646b2`.

PR #447, opened 2015-12-28
Some refactoring the common project.

zaneli:
* Removed unused imports.
* Replaced Scalaz [`node`](https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Tree.scala#L233-L234), [`leaf`](https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Tree.scala#L237-L238) methods.
* Converted match statement to pattern matching anonymous function.

etc...

### DIFF
--- a/common/src/main/scala/org/specs2/control/Status.scala
+++ b/common/src/main/scala/org/specs2/control/Status.scala
@@ -35,7 +35,7 @@ sealed trait Status[+A] {
                                fail: String => X,
                                exception: Throwable => X,
                                both: (String, Throwable) => X): X =
-    fold(ok, _ match {
+    fold(ok, {
       case This(m) => fail(m)
       case That(e) => exception(e)
       case Both(m, e) => both(m, e)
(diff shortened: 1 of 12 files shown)


etorreborre:
Thanks for the PR. I agree with almost everything but the replacement of `node` and `leaf`. It is in general a good practice to hide constructors, to only be left with`Tree` instead of the more specific `Node`  or `Leaf` type. This helps with type inference and can sometimes be a way to refactor the ADT without modifying the API.

zaneli:
I see, Thank you.

But [`Tree.apply`](https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Tree.scala#L191-L192) seems to be deprecated, too.
And this is arity-1 as alias `Leaf.apply`, so I'd like to replace [`Node.apply`](https://github.com/etorreborre/specs2/pull/447/files#diff-9a85f1a5158417a1baa3a95e71b3ad24R41), [`Node.apply`](https://github.com/etorreborre/specs2/pull/447/files#diff-9a85f1a5158417a1baa3a95e71b3ad24R53) with `Tree.apply`, but I couldn't...

Could you give me some advice what I should do about it?

etorreborre:
Oh sorry, I didn't realize those changes in the API. This is all fine then.

zaneli:
Thanks!

Issue #458, opened 2016-03-03
"must not(be)" doesn't fail


holograph:
The following assertion passes, but should actually fail:

```scala
        val a = 8
        a must not(be)
```

This might seem a bit strange, but using `be` is a useful pattern in matcher libraries, for example:
```scala
case class Result(...)
case class ServerResponse(results: Seq[Result], ...)

def haveResults(resultsMatcher: Matcher[Seq[Result]] = be): Matcher[ServerResponse] =
  resultsMatcher ^^ { resp: ServerResponse => resp.results }

val someResult: Result = ...
serverResponse1 must haveResults(contain(someResult))
serverResponse2 must not(haveResults())     // <-- This fails

**Comment by etorreborre** [comment]
Hi Tomer, I think `not(beEmpty)` should be used here:
```scala
class TestMutableSpec extends mutable.Specification with ResultMatchers {

  case class Result(n: Int)
  case class ServerResponse(results: Seq[Result])

  def haveResults: Matcher[ServerResponse] =
    haveResults(not(beEmpty[Seq[Result]]).updateMessage(_ => "There are no results"))

  def haveResults(resultsMatcher: Matcher[Seq[Result]]) =
    resultsMatcher ^^ { resp: ServerResponse => resp.results }

  val someResult: Result = Result(3)

  eg { ServerResponse(List(someResult)) must haveResults(contain(someResult)) }
  eg { ServerResponse(List(someResult)) must haveResults }
  eg { ServerResponse(List()) must not(haveResults) }

  eg { (ServerResponse(List(someResult)) must not(haveResults)) must beFailing }
  eg { (ServerResponse(List()) must haveResults) must beFailing }
}
```

All those examples pass as expected
```
[info] TestMutableSpec
[info]
[info] + ServerResponse(List(someResult)) must haveResults(contain(someResult))
[info] + ServerResponse(List(someResult)) must haveResults
[info] + ServerResponse(List()) must not(haveResults)
[info] + (ServerResponse(List(someResult)) must not(haveResults)) must beFailing
[info] + (ServerResponse(List()) must haveResults) must beFailing
[info]
```

Also note the overloading on `haveResults` which allows to drop the parentheses :-).

holograph:
[...]

PR #460, opened 2016-03-08
fix showOnly flag to exclude title

zaneli:
fix this error (might be from https://github.com/etorreborre/specs2/commit/7f9efd99fccbc465fa3481911824790170e58ccf).

```
[error] x the title is not displayed if # can not be shown
[error]  '[info]_title' contains '[info]_title' (file:1)

[error] Failed: Total 719, Failed 1, Errors 0, Passed 718
[error] Failed tests:
[error] 	org.specs2.reporter.TextPrinterSpec
```

### DIFF
--- a/core/src/test/scala/org/specs2/reporter/TextPrinterSpec.scala
+++ b/core/src/test/scala/org/specs2/reporter/TextPrinterSpec.scala
@@ -1,14 +1,14 @@
 package org.specs2
 package reporter
 
-import matcher.{MustMatchers}
+import matcher.MustMatchers
 import specification._
-import dsl.{FragmentsDsl}
-import specification.create.{DefaultFragmentFactory}
+import dsl.FragmentsDsl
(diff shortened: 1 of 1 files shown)


zaneli:
Sorry...this modification might have been occurred some other errors.
I'll try fix it.

zaneli:
I have confirmed that all build jobs succeeded in my repository...
https://travis-ci.org/zaneli/specs2/builds/114425085

I'll try to check a little more.

etorreborre:
Sorry about that, I fixed this test now. The implementation of the test was wrong but was not showing up because the bug in #458 was hiding it.

etorreborre:
Actually your implementation is better (because it is testing with all other flags) so I merged it in a different commit (and attributed it to you). Thanks!

zaneli:
I see.
Thank you, too!

PR #471, opened 2016-04-16
Configure external links in ScalaDoc

Atry:


etorreborre:
This is a good idea, thanks. However I dont think that I have external links anywhere in the doc. Maybe it is time to add some?

Atry:
There are some signatures of methods that reference Scala standard library.

PR #476, opened 2016-04-25
fix dependency settings

xuwei-k:
I think the common module should not depends on kind-projector.
Because it is a compiler-plugin, not library. unnecessary in runtime.


etorreborre:
Thanks for this.

Issue #462, opened 2016-03-24
Inconsistent behaviour of eventually in EventuallyMatchers and EventuallyResults


jcazevedo:
The behaviour of `eventually` is inconsistent in the `EventuallyMatchers` and the `EventuallyResults` traits. Take the following example into account (where I make use of `println` for easier debugging):

```scala
def f = {
  println("here")
  false
}

eventually(retries = 1, sleep = 100.milliseconds) {
  f must beTrue
}

f must beTrue.eventually(retries = 1, sleep = 100.milliseconds)
```

The first `eventually` calls `f` twice, whereas the second one only calls it once. I believe this has to do with how specs2 handles the conversion from a `MatchResult` to `Result`. The body of the `retry` function inside `eventually` in the `EventuallyResults` trait is the following:

```scala
lazy val t = r
val result = ResultExecution.execute(t)(AsResult(_))
if (result.isSuccess || retries == 1)
  t
else {
  Thread.sleep(sleep.toMillis)
  retry(retries - 1, sleep, r)
}
```

The first conversion of `r` to result will raise an exception in case the match result is a failure and `t` won't be set, which results in a second evaluation after the `if`.

etorreborre:
I need to have a look at that but the first behaviour the intention of `retries`, that is, the number of times you want to retry something if it doesn't succeed the first time. So the question for me is more: why does it execute `f` only once when used with a matcher? I don't have time right but I'll take a look during the week-end. Thanks for reporting.

jcazevedo:
> I need to have a look at that but the first behaviour the intention of retries, that is, the number of times you want to retry something if it doesn't succeed the first time.

At first I also interpreted `retries` as such, but looking at the code it seems to be implemented as the number of tries. That's seems like something easier to fix than the issue I've raised, though.

etorreborre:
This is fixed now (damned exceptions) and you were right about the meaning of `retries`, it is the total number of tries and not what I thought it should be. I am not going to change this for now in order to avoid breaking too many specs in the wild.

etorreborre:
The fix is in `3.7.2-20160326084340-6b923de`.

PR #479, opened 2016-05-05
Make EventuallyResults consistent with EventuallyMatchers

luismfonseca:
This changes the API so that the eventually returns again.
This will make specs that were built under this assumption continue to work as expected.

The following example used to return an error prior the commit 471cd6c8ad97d, but now it will pass:
```
"example" in {
  eventually {
    ko
  } // previously we weren't required to do a `must beSuccessful` here
  ok
}
```
This means that all existing specs using `eventually` this way, aren't currently able to see the error and the test will always pass.

The previous issue #462 was kept resolved.


etorreborre:
This looks good to me. I think there was a reason for me returning `Result` instead of `T` but I can't remember now.

PR #483, opened 2016-05-24
Fix bug in regex matching with multiline strings

thomaslloyd:


etorreborre:
Thanks for the fix! I'll try to publish an interim version very soon for you.

etorreborre:
It will be available as `3.8.3-20160524134053-8145e17`.

thomaslloyd:
Awesome! 👍

Issue #341, opened 2015-03-08
BeEqualTo should support deeper compare


noam-almog:
Hi,

Specs2 output in case of equality is lacking if you are matching two objects.

You will get a full output that says 

```
Expected: ClassA(val1.. valN) 
Actual: ClassA(val1'... valN')
```

without you actually knowing what is the difference. In case of random values and more than two properties it gets really hard to actually know what the difference.

I want to PR specs2 with a change that will give you a more meaningful response with the actual properties that are different.

For example:

```
Expected: ClassA(val1,...valN)
Actual: ClassA(val1,...valN)
Differences:
property [val1] Expected: someVal, Actual: anotherVal
property[valX] Expected: someVal, Actual: anotherVal
```

I'm open to suggestions about the format.
- Noam

etorreborre:
I think that we should restrict this to case classes, so probably to classes extending `Product` but we need to make sure that this doesn't clash with things like `Tuples`.

Then if we do a deep comparison we need to make sure that we display failures only for the leaves and not at each level. Other than this I guess that we can reuse most of the existing machinery for comparisons (on lists if your attribute is a `List` etc...).

Probably we can add a new `Details` subclass (see `Result.scala`): `ProductDetails` to hold the differences and extend the `Diffs` trait to handle this case (and do a default implementation in `SmartDiffs`).

andriuss:
:thumbsup:

mdedetrich:
I wouldn't recommend adjusting `beEqualTo` to use anything other than the standard `==` check (and if you want to do deep equality checking, should probably make a `beDeepEqualTo` or something like that).

Many people still use `==` to test for equality, and `beEqualTo` is a very handy method to test if an `override equals(a: Any) actually works as expected.

etorreborre:
I should probably provide a new set of matchers for `scalaz` and for `cats` using their respective `Eq` typeclasses.

etorreborre:
This issue would be best solved by using a matcher delegating the diff to https://github.com/xdotai/diff. I have sketched such a matcher [there](https://gist.github.com/etorreborre/5699231208c47ee78d6eee0f015e5a1b) (but got some error I don't yet understand).

noam-almog:
@etorreborre 
I'm starting to work on a PR for this issue and I wanted to get some feedback. 
I will try to incorporate some of the ideas from xdotai's code.

The basic outline of this would be:

Primitives:
left -> right, for example: 1 -> 2

Traversable:
[...]

PR #500, opened 2016-08-18
BeEqualTo should support deeper compare

noam-almog:
Hi Eric,

Decided to let you in on the current progress and get some feedback.
This is an implementation of issue #341 

Error messages are:
Primitives: 1 != 2  and 'a' != 'b'

Scala Collections:
   Sets: Set('a', 'b', added: 'c', 'd', removed: 'e')
   Seq: Seq('a', 'b' != 'c', added: 'd', 'e', removed: 'f')
   Map: Map('a' -> 'b', 'c' -> {'d' != 'x'}, added: 'g' -> 'h', removed: 'e' -> 'f')

Scala:
   Option: Some('abc' != 'def'), None ==> Some(...)
   Either: Right('abc' != 'def'), Right(...) ==> Left(...)

Case Class:
   Some case class: ClassName(prop: 'val1' != 'val2', prop2: 'val')

Right now the code is not attached to the BeEqualTo matcher because it's using BeEqualTo on the tests that tests the code.


etorreborre:
Great. I am leaving today for one week of vacation but I will have a look at that when I come back.

noam-almog:
@etorreborre any chance you had some time to look at it. This PR will require some lower level changes that i don't want to get into without your support.

etorreborre:
I just came back today to work and have plenty of things to catch up with. I will have a look at your PR tomorrow morning.

etorreborre:
@noam-almog I had a closer look at your PR. That's an impressive piece of code!

For now I have 2 questions (which I am also asking myself):

 1. can we remove the dependency on Shapeless? 

I have the long-term goal to remove all dependencies for `specs2-core` and I am currently working on removing the internal `scalaz-stream` copy (next is scalaz). If we absolutely need it for case class diffs we might have to put it in a different module. Otherwise there will be conflicts with different shapeless versions in user projects.

 2. how do you see the integration with `beTypedEqualTo` work?

I am a bit reluctant to change too much the internals or the API of what's already there because I think at this stage stability is quite important. 

One approach would be to piggy-back on the [`Details` trait](https://github.com/etorreborre/specs2/blob/master/common/src/main/scala/org/specs2/execute/Result.scala#L340) (modifying `FailureDetails` to take in `Any, Any` instead of `String, String`) then  "inject" your diffs in the [`SmartDiffs`](https://github.com/etorreborre/specs2/blob/master/common/src/main/scala/org/specs2/main/Diffs.scala#L35) trait, using reflection to know which case to apply. This could work for common cases but not automatically (meaning "just by defining a typeclass instance") for user-defined types.

Another approach would be to have a separate matcher doing nice diffs and requiring a `DifferenceInspector` instance. Actually if we are going this route we should as well add a `Equality` instance to allow `Doubles` inside a case class to be compared with some degree of precision only. Question: how to name that new thing?

A variant of option 2 is to evolve the existing `BeTypedEqualTo` class and add the missing pieces for nice diffs but we need to check if that can be done without breaking any user code.

3. I would like to integrate 2 other pieces to the "diff" thing (done in the `SmartDiffs` class at the moment)

- shortened diffs: if you have differences on strings where just one letter differs it's nice to see it right away

- sequence diffs: there is an attempt at "best matching" sequences to find the least number of different elements between 2 sequences. For example if you have just one element inserted in the middle of the expected sequence it is nice to be able to see it right away instead of seeing many elements "removed" and "added"

noam-almog:
@etorreborre 

The goal to leave a smaller footprint is a great goal and I fully support it.
The problem is that shapeless is the missing peace in the big puzzle here which is to move comparisons from runtime to compile time. The current implementation gets the proper inspector on compile time using scala implicit resolving mechanism, and the internal comparison of the case class is done on compile time using shapeless. The problem I have with removing shapeless is that it will force us to change the approach from compile time to runtime. This will actually kill the current design of scala implicit resolution.

Future Integration:
Basically the biggest difference is that we want to be able to propagate types from user code to matcher code on compile time, the only way is to get the implicit inspector on the user test call site. To safely do that we can create a different path in the code, beEqualTo2 :) or beEqualToExperimental allowing users to play with it and later on replace the old implementation or just leave both of them.

Smarter diffs is a feedback I got from many different people that I talked with. With the current design those issues can be added by just modifying the String comparator (or Traversable) individually. To me, improving diffs is great but we need to sort out what is next.
Should we break the compile time and revert to runtime or should we continue with integrating this code to be a proper matcher and then improve.
So what do you think we should do ?

etorreborre:
I think it is worth integrating most of the compile-time stuff to `BeTypedEqual` but we need to leave the case class instances to a separate module which requires shapeless (let call it `specs2-generic` for example). Otherwise we risk breaking the full core functionality for one feature.

noam-almog:
@etorreborre I'm not sure i understand what you mean by specs-core module, if those are the project modules then this whole code sits on the matcher module.

etorreborre:
`specs2-core` is the minimum specs2 jar needed to write specifications. It depends on `specs2-common` and `specs2-matcher`. If we add more dependencies to `specs2-matcher` we will get them when using `specs2-core`.

noam-almog:
make sense, the main issue is that with this design you are actually allowing user to implement his own inspector implementations but providing all of the default ones so he won't have to.
So even if we extract this code to another module we still need to define the implicit reference to it on matcher module which will drag it back into the core.

etorreborre:
What if we just leave [this](https://github.com/etorreborre/specs2/pull/500/files#diff-f3ba22632d788ef0ba911db597487c47R52) in a separate module to be imported with an object or a trait?

noam-almog:
yes, LowerPriority is mixed in with DifferenceInspectable object  so it will drag it back in.

noam-almog:
@etorreborre I've replaced all current equal matcher to point to the new impl (type and not type)
can you please take a look ?

etorreborre:
I will but probably only tomorrow, thanks.

noam-almog:
10x, since it's a big change, i'm trying to get as much feedback as I can

matcher-extra/src/main/scala-scalaz-7.0.x/org/specs2/matcher/ProcessMatchers.scala
```diff
@@ -34,7 +36,7 @@ trait ProcessMatchers extends Expectations {
       result(r, e)
     }
 
-    def withValues(values: Seq[T]): ProcessMatcher[T] =
+    def withValues(values: Seq[T])(implicit di: DifferenceInspectable[T]): ProcessMatcher[T] =
```
etorreborre: One thing about the name. I would rather have a shorter name for this typeclass, like `Diffable`. I have been trying to keep short names in specs2 as much as possible.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
```diff
@@ -0,0 +1,76 @@
+package org.specs2.matcher
+
+import org.specs2.execute._
+import org.specs2.matcher.describe.{DefaultDifferenceInspector, DifferenceInspectable, DifferenceInspector}
+import org.specs2.text.NotNullStrings._
+
+import scala.collection.{GenTraversable, GenTraversableOnce}
+
+
+class BeTypedEqualTo2[T : DifferenceInspectable](t: =>T, equality: (T, T) => Boolean = (t1:T, t2:T) => t1 == t2) extends AdaptableMatcher[T] {
```
etorreborre: Why did you create a `BeTypedEqualTo2` class instead of modifying `BeTypedEqualTo`? I'm pretty sure there is code out there using that class directly so we should better leave it and just replace it with your code for `BeTypedEqualTo2`.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
```diff
@@ -0,0 +1,76 @@
...
+  outer =>
+
+  protected val ok: String => String = identity
+  protected val ko: String => String = identity
+
+  protected val differenceInspector: DifferenceInspector = new DefaultDifferenceInspector
+
+  def adapt(f: T => T, okFunction: String => String, koFunction: String => String) = {
+    new BeTypedEqualTo(f(t), equality) {
```
etorreborre: I see you are doing some delegation to the "old" class here. I suppose that you could have as well written `new BeTypedEqualTo2` here, right?

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
```diff
@@ -0,0 +1,76 @@
...
+  def apply[S <: T](b: Expectable[S]): MatchResult[S] = {
+    val (actual, expected) = (b.value, t)
+    val isEqual = checkEquality(actual, expected)
+    val diff = differenceInspector.diff(actual, expected)
+
+    failureDetailsFor(actual, expected)
+      .map( d =>
+        result(isEqual,
+          ok(s"${b.description} is equal to '${expected.notNull}'"),
+          ko(b.describe(diff.render)), b, d)
```
etorreborre: I see that you have changed failure messages to be `!=` instead of `is not equal to`, which is particularly better for case classes when you have lots of different messages to display for different fields. Maybe we should do the same for the `is equal to` message and replace it with ` == ` for symmetry.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @return a MatchResult describing the outcome of the match
+   */
+  def apply[S <: T](b: Expectable[S]): MatchResult[S] = {
+    val (actual, expected) = (b.value, t)
+    val isEqual = checkEquality(actual, expected)
+    val diff = differenceInspector.diff(actual, expected)
+
+    failureDetailsFor(actual, expected)
+      .map( d =>
```
etorreborre: Stylistic comment. Sometimes you leave some space before / after parentheses. Can you please reformat the code to make it consistent with the rest of specs2 where there is no such space?

matcher/src/main/scala/org/specs2/matcher/OptionMatchers.scala
```diff
@@ -15,7 +16,7 @@ private[specs2]
 trait OptionBaseMatchers {
 
   def beSome[T](check: ValueCheck[T]): SomeCheckedMatcher[T] = SomeCheckedMatcher(check)
-  def some[T](t: T): SomeCheckedMatcher[T] = beSome(t)
+  def some[T : DifferenceInspectable](t: T): SomeCheckedMatcher[T] = beSome(ValueChecks.valueIsTypedValueCheck(t))
```
etorreborre: Out of curiosity. The code `beSome(t)` used to compile like that. Did you have to add `ValueChecks.valueIsTypedValueCheck(t)` to make it compile or can the code compile as before? I am wondering if that means there could be regressions in type inference and implicits resolution because of the introduction of a new typeclass parameter.

matcher/src/main/scala/org/specs2/matcher/describe/DifferenceInspectable.scala
```diff
@@ -0,0 +1,60 @@
...
+  implicit def mapInspector[K : DifferenceInspectable, V : DifferenceInspectable]: DifferenceInspectable[Map[K, V]] = new MapDifferenceInspectable[K, V]
+  implicit def setInspector[E: DifferenceInspectable]: DifferenceInspectable[Set[E]] = new SetDifferenceInspectable
+  implicit def seqInspector[E: DifferenceInspectable]: DifferenceInspectable[Seq[E]] = new SeqDifferenceInspectable
+  implicit def arrayInspector[E: DifferenceInspectable]: DifferenceInspectable[Array[E]] = new ArrayDifferenceInspectable
+}
+
+trait DifferenceInspectableIntrospectors extends Fallback {
+//  implicit def traversableInspector[E: DifferenceInspectable]: DifferenceInspectable[Traversable[E]] = new TraversableDifferenceInspectable[E]
+
+  implicit def caseClassInspector[T <: Product with Serializable: IsCaseClass, L <: HList](implicit labelled: LabelledGeneric.Aux[T, L],
```
etorreborre: This is still a big issue to me. I know that the major value of this PR is to have nice case class diffs but having to pull-in shapeless as a dependency will just make the whole project unusable for many users. I see 2 ways out:

 - create a separate module providing additional implicit imports to diff case classes
 - bite the bullet and recode the subset of shapeless necessary to walk through case classes (and put that as a package in `specs2-common`). Eventually I will do exactly this for all the other dependencies, from scalaz-stream to Scalaz.

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
```diff
@@ -0,0 +1,192 @@
...
+
+class EitherRightDifferenceInspectable[R : DifferenceInspectable](implicit rdi: DifferenceInspectable[R])
+  extends DifferenceInspectable[Right[Nothing, R]] {
+  def diff(actual: Right[Nothing, R], expected: Right[Nothing, R]) = (actual, expected) match {
+    case (Right(a), Right(e)) if a == e => EitherIdentical(rdi.diff(a, e), isRight = true)
+    case (Right(a), Right(e)) if a != e => EitherDifferent(rdi.diff(a, e), isRight = true)
+  }
+}
+
+class EitherLeftDifferenceInspectable[L : DifferenceInspectable](implicit ldi: DifferenceInspectable[L])
```
etorreborre: Can you also please add a disjunction `DifferenceInspectable` for Scalaz `\/` and a also one for `cats.data.Xor` in the `specs2-cats` module?

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private def checkEquality(actual: Any, expected: Any): Boolean =
+    (actual, expected) match {
+      case (e1: Array[_], e2: Array[_]) => e1.deep == e2.deep
+      case (e1, e2) => e1 == e2
+    }
+
+  private def failureDetailsFor(actual: Any, expected: Any): Option[Details] =
+    (actual, expected) match {
```
etorreborre: Just taking a step back here, this means that we have both:
 
 - your new diffs as regular failure messages
 - some additional details provided by my `SmartDiffs` object

I wonder what this looks like when running a spec like:
```scala
class MySpec extends Specification { def is = s2"""
   test ${List(1, 2, 3) ==== List(2, 3, 4)}
"""
}```

File: tests/src/test/scala/org/specs2/matcher/AnyMatchersSpec.scala [hunk 10]
```diff
@@ -106,7 +106,7 @@ Implicits
   + the must implicits can be deactivated with the NoMustExpectations trait
                                                                                                                         """
   "be like" - new group {
-    eg := { (List(1, 2) must beLike { case List(a, b) => (a + b) must_== 2 }) returns "'3' is not equal to '2'" }
+    eg := { (List(1, 2) must beLike { case List(a, b) => (a + b) must_== 2 }) returns "'3 != 2'" }
```
etorreborre: I don't understand why the message is quoted with `'` here. Is that really useful?

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
```diff
@@ -34,20 +40,27 @@ class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers
...
+     """Array(1, 3 != 2)""" }
 
   Set equality
-  ${ (Set(1) must_== Set.empty[Int]) returns "Set(1) is not equal to Set()"}
+  ${ (Set(1) must_== Set.empty[Int]) returns "'Set(1) != Set()'"}
   ${ (Set(1, 2) must be_==(Set(2, 3))) returns
-      """Set(1, 2) is not equal to Set(2, 3)""" }
+      """'Set(1, 2) != Set(2, 3)'""" }
+  ${ (Set(1, 2) must be_===(Set(2, 3))) returns
+      """Set(2, added: 3, removed: 1)""" }
```
etorreborre: I wonder how the diff looks like when there are lots of elements. If there are lots of `added` and `removed` labels I am not sure how readable this will be. Should we:

 - have smaller labels: `[+]` and `[-]`?
 - have a recap:

```
added:
  3
removed:
  1
```

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
noam-almog: I had to do this in stages due to the fact the specs is testing itself using beEqualsTo matcher.
So I created a new one and moved all usages to it.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
noam-almog: i didn't notice this, it's a mistake.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
```diff
@@ -34,20 +40,27 @@ class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers
...
+      """Map(1 -> {2 != 4}, 3 -> {4 != 1})""" }
 
   Other collections use normal equality but display missing elements
   ${ Seq(1, 2) must be_==(Seq(1, 2)) }
   ${ (Seq(1, 2) must be_==(Seq(2, 3))) returns
-     """List(1, 2) is not equal to List(2, 3)""" }
+     """'List(1, 2) != List(2, 3)'""" }
+  ${ Seq(1, 2) must be_===(Seq(1, 2)) }
+  ${ Seq(1, 2) must be_===(Seq(2, 3)) returns
+     """Seq(1 != 2, 2 != 3)""" }
```
etorreborre: I am not sure that rendering differences like that for sequences work well when you have an additional unexpected element in your sequence because then you see lots of `!=` (with possibly large strings representing the different elements on each side of `!=`) where there is actually just one difference, the element being added to the sequence.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
```diff
@@ -34,20 +40,27 @@ class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers
...
 
   Expected values are kept in the failure details
```
etorreborre: Github does not help the review with all the red highlighting... I think I had issues like that in the past where there was indeed a problem in my text. Do you think this is a bug in Github or this is something which you could arrange?

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
noam-almog: This part is up to you, if you ask me, it's better to have just one consistent diff in the system. Once the new diff will prove itself useful to developers I think it would be smart to pull it to the core instead of the old one.

tests/src/test/scala/org/specs2/matcher/describe/DifferenceInspectorSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+    different Seqs should show which elements were changed       ${ differenceDescriber.diff(Seq("a"), Seq("b")) must_== SeqDifference(Seq(PrimitiveDifference("a", "b")), Seq.empty, Seq.empty)}
+    different Seqs should show which elements were added         ${ differenceDescriber.diff(Seq("a"), Seq("a", "b")) must_== SeqDifference(Seq(PrimitiveIdentical("a")), Seq("b"), Seq.empty)}
+    different Seqs should show which elements were removed       ${ differenceDescriber.diff(Seq("a", "b"), Seq("a")) must_== SeqDifference(Seq(PrimitiveIdentical("a")), Seq.empty, Seq("b"))}
+    different Seqs should show changed and added with null       ${ differenceDescriber.diff(Seq("a"), Seq(null)) must_== SeqDifference(Seq(PrimitiveDifference("a", null)), Seq.empty, Seq.empty)}
+    be able to compare Seq[Any]                                  {differenceDescriber.diff(Seq("a", 1, 2l, 3.3d, 4.4f), Seq("b", 2, 3l, 4.4d, 5.5f)) must_== SeqDifference(Seq(PrimitiveDifference("a", "b"), PrimitiveDifference(1, 2), PrimitiveDifference(3.3d, 4.4d), PrimitiveDifference(4.4f, 5.5f)), Seq.empty, Seq.empty)}
```
etorreborre: It's missing a `$` here.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
```diff
@@ -21,9 +21,15 @@ class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers
...
   ${ "a" must be_===("a") }
+  ${ "a" ==== "a" }
+  ${ "a" must not(be_===("b")) }
+  ${ "a" must be_!==("b") }
+  ${ "a" must_=== "a" }
+  ${ "a" must_!== "b" }
   // doesn't compile
   // { "a" ==== 1 }
   ${ "a" must not be_===("b") }
+  { "a" must not be_====("b") } todo: problem with not matcher
```
etorreborre: I would like to understand that one. The `not` business in specs2 is a bit tricky.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
etorreborre: Ok, it makes sense, I know that circular feeling :-).

matcher/src/main/scala/org/specs2/matcher/OptionMatchers.scala
noam-almog: it used to be resolved implicitly. I played with various code changes to this and in the end I just made it consistent across a few call sites. From what i saw it didn't have any issues.
I did notice that the not matcher wipes types, so if you do 'not be_===' types are gone and we revert to runtime resolution.

matcher/src/main/scala/org/specs2/matcher/describe/DifferenceInspectable.scala
noam-almog: Using shapeless is not related to the nice diffs, there is other code that does that. 
shapeless allows me to 
1. do stuff at compile time vs runtime 
2. it reduces the boilerplate of resolving a Diffarable.
3. Allows us to inject new differables easily, kind of an open close principle.

The main question is how much shapeless is a problematic library. most of the developers don't use it. some libraries do and from what i see it's pretty stable. while scalaz or akka enforces things on the project, i don't thing shapeless is at the same category.

tests/src/test/scala/org/specs2/matcher/AnyMatchersSpec.scala
noam-almog: due to the way Expectable.describe works, it gets the description and quote it in some cases.
I didn't want to touch that just yet because it's causing too many tests to fail.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: when you do, let me know, it's on my todo list for this PR

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: This is exactly why i moved all of the render code to a different class.
we can play with this however you want. I didn't have guard rails for very long values yet, but I think that would be a great addition.
BTW, my problem with this PR is that if === operator will display a very good output it will make people use it more instead of writing custom matchers. when you compare realy big objects i think that you are doing something wrong, but that is just my opinion.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: you mentioned this on an early comment, i totally agree, question is if we want to do that now.
specs2 does not support this type of comparison right now. don't you think it's something that can be handled on another PR ?

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: first time that i see something like this...

matcher-extra/src/main/scala-scalaz-7.0.x/org/specs2/matcher/ProcessMatchers.scala
noam-almog: All comments about naming are more than welcome. I'm not attached to any of the names here.

tests/src/test/scala/org/specs2/matcher/describe/DifferenceInspectorSpec.scala
noam-almog: it doesn't, i have some issues with this, will fix it soon.

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
noam-almog: in order for them to work we need to have the developer importing them at the call site
```
   import ScalazDiffables._
```
Otherwise it will compare stuff with the fallback diffable

matcher/src/main/scala/org/specs2/matcher/describe/DifferenceInspectable.scala
etorreborre: Sorry I still haven't had time to review all your comments, but I wanted to add a note on the question of having dependencies for specs2. One issue with dependencies is that I cannot build a complete release of `core` if I don't have all my dependencies for a given version. For example I wanted to release specs2 for `2.12.0-RC1` today but shapeless is only available for `2.12.0-M5`. Having to wait for a cascade of dependencies when a new version of Scala is out is not really good, especially if lots of people are waiting for specs2 to build their own version.

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
etorreborre: I wonder if we could use [export-hook](https://github.com/milessabin/export-hook) for this. The only issues are:

 - it doesn't work with Scala 2.12
 - it depends on shapeless (again...)

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
etorreborre: I agree with you. Writing custom matchers is a good thing to do in many situations. Except when dealing with structured data. In that case you would have to write lots of nested custom matchers and a good diff is a lot better.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
etorreborre: I realize that at the moment this is done by the `SmartDiffs.showSeqDiffs` method which is called in the `TextPrinter` for example. So since this PR still creates `FailureSeqDetails` objects when there is a mismatch, the `added/missing` lines should still be reported in the console.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
etorreborre: If I remember correctly it was tricky to type `not` and still be able to have a syntax like `a must not beCorrect`. I am not sure we can really fix it :-(.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: right now i didn't touch the Actual:, Expected: part, it will remain the same as it was.
The message is being set by the matcher and it's using my renderers.
In the future I think color coding is a great option but I didn't want to do that right now.

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
noam-almog: so in order for us to remove shapeless we have to include shapeless :)

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
noam-almog: resolved.

matcher/src/main/scala/org/specs2/matcher/BeTypedEqualTo2.scala
noam-almog: resolved

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
noam-almog: I will consult with some people about this, I need to have a way to resolve those implicits once someone includes the cats and scalaz matchers.
Probably the next big thing i need to handle.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: a comment about this, i still prefer added and remove labels instead of [+] and [-] which are less intuitive to me. let me know if you still want this change.

tests/src/test/scala/org/specs2/matcher/describe/DifferenceInspectorSpec.scala
noam-almog: fixed.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: due to the fact the 'not' wipes the type chain right now if you are doing not XXXX then you are basically getting the Fallback comparator. The issue is resolved but when the matcher fails it will give out less feedback to the user.

matcher/src/main/scala/org/specs2/matcher/describe/inspectable/inspectable.scala
noam-almog: @etorreborre just played with the xor matcher and it seems like it's already being handled.
```
${ Right(Seq(1, 2, 3)) must beXorRight(Seq(1, 2)) }
```
```
Right(List(1, 2, 3)) is Right but Seq(1, 2, removed: 3)
```
What is happening that everything that gets comparisons from ValueCheck.
I struggled a bit adding the Diffable implicit to the XorMatchers but from what i see i don't really need to fix anything there right now.

tests/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
noam-almog: @etorreborre it seems like github fixed their bug :D

noam-almog:
@etorreborre I've completed this PR, all tests are passing and I changed the message formatting like we discussed. There might be some things you would probably want to change.
There is a lot of code to look at, and I want to work with you to merge this PR back to master, when will you have time to review this ?

etorreborre:
Hi @noam-almog, sorry I missed your message. I will try to take some time to review the PR this week. My main concern I think is the dependency on shapeless. I am currently working hard to remove scalaz-stream, then scalaz from the equation and I wouldn't like to have to maintain future cross-releases for different shapeless versions. 

I would like to explore the use of [`export-hook`](https://github.com/milessabin/export-hook#the-type-class-provider) to allow a minimal dependency but still keep the possibility to get case class `Diffable` instances discovered without explicit imports. 

Are you ok for trying this option?

noam-almog:
@etorreborre so the call site will have to include this ?
I would have to say that this would not be my favourite option but I can live with that.
I can only say that shapeless is not such a popular library like scalaz and with the post beta releases from scala it won't cause too much havoc.
Another thing is, to remove it later on from the default Diffable is very simple, if it will cause more issues with users than it solved moving it out of the core is very simple, we can try and give it a chance and see how it goes before preparing for the worst option, WDYT ?

etorreborre:
The problem is not so much that we need to wait for a new Shapeless version when there is a new Scala version but the fact that 2 versions of shapeless might not be binary compatible. And if a user needs a given version for the production code it might conflict with specs2.

etorreborre:
Hi @noam-almog I got another idea. If we put the shapeless implicits for dealing with case classes in a `specs2-shapeless` module and more precisely in the `org.specs2` package object.

Then anyone adding the `specs2-shapeless` dependency to the build will get those implicits pretty much transparently. What do you think?

noam-almog:
we can also repackage shapeless or at least the parts we are using

etorreborre:
I just tried my suggestion and nothing seems to work. I will try`export-hook` next. I am also not a big fan of inlining the `LabelledGeneric` code into specs2 because it is quite complex. Managing dependencies is hard :-(

etorreborre:
@noam-almog I gave it one more try, rebased on master and was finally able to pass all the tests. For now I am staying very conservative with the shapeless dependency. This means that getting nice diffs for case classes will require an import of `org.specs2.matcher.CaseClassDiffs` or mixing the same trait (and depend on the new `specs2-shapeless`). This is not as nice as getting it out of the box but I don't know how to do it differently for now. I will publish soon.

noam-almog:
@etorreborre still, great news looking forward to get some feedback about this.
thanks for not giving up on this PR 👍

PR #506, opened 2016-09-13
adding generics to BaseMatcher in HamcrestMatcherAdapter

nadavwe:
hey @etorreborre,
I'm trying to use the HamcrestMathcerAdapter (since i like specs2 matchers much more...) but i had some trouble doing that since this case class does not extend Matcher[T].
was the `[T]` left out by mistake or in purpose?
Thanks!


etorreborre:
That looks good to me. I think I was not extending `BaseMatcher[T]` because that type parameter was not available in previous versions of Hamcrest. Thanks. If you wait a bit I'll publish a new version with this in.

etorreborre:
The fix is in `3.8.5-20160913153617-aacd4a1` if you want to try it.

nadavwe:
@etorreborre thanks!

PR #508, opened 2016-09-14
Mockito project has moved to GitHub

emma-burrows:
Just correcting a link since the Mockito Google Code page says to look at GitHub instead.


etorreborre:
Thanks!

PR #510, opened 2016-09-14
describeTo works to show the expected result

nadavwe:
hey @etorreborre,
continuing my HamcrestMatcherAdapter saga... =)
so it appears that the describeTo uses its description to show the expected result.
the current implementation creates weird results. 
for example:

``` scala
  @Test
  def hamcrest = {
    val matcher: Matcher[Boolean] = beTrue
    assertThat(false, matcher) // there's an implicit conversion here, but it just uses the adapter
  }
```

outputs:

```
Expected: the value is false
     but: was <false>
```

which is quite confusing...

i don't think my change is the best, i'm just not sure how to solve this...

thanks,
Nadav

### DIFF
--- a/mock/src/main/scala/org/specs2/mock/HamcrestMatcherAdapter.scala
+++ b/mock/src/main/scala/org/specs2/mock/HamcrestMatcherAdapter.scala
@@ -22,7 +22,7 @@ case class HamcrestMatcherAdapter[T](m: Matcher[T]) extends BaseMatcher[T] {
     val i = if (item != null && item.isInstanceOf[Function0[_]]) item.asInstanceOf[Function0[_]].apply().asInstanceOf[A] else item
     try {
       matcher.apply(Expectable(i)) match {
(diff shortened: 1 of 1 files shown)


etorreborre:
I need to do a bit of testing around this to make sure the display is ok both in your case and with Mockito. I'll get back to you.

etorreborre:
Sorry for the "I'll get back to you"... I remember trying out your change in some other cases and the results were looking weird as well. I am bit swamped at the moment with other things, I hope I'll find some time to come back to this for real.

nadavwe:
no problem. 
i'm not waiting for it, just stumbled upon and thought it's something worth mentioning.
as i said, I don't think my solution is very good... probably need a more comprehensive solution.

etorreborre:
I am going to close this for now and put it in the "known issues" bucket. It is a bit hard to reconcile the hamcrest and the specs2 view on failure messages.

PR #513, opened 2016-10-04
make version-specific source directories work in community build

SethTisue:
I'm trying to move the Scala 2.12 community build to use specs2's
master branch, rather than a fork.

in the context of the community build, the Scala version number
(including the scalaBinaryVersion) is liable to be something like
"Scala 2.12.0-1f6006d-nightly", so anything that inspects the Scala
version should be robust about inspecting only the beginning.

these long Scala version numbers also break the trick you're doing of
using the Scala binary version verbatim as the name of a
Scala-version-specific directory.

so this commit changes that logic in a way that works for the
community build.  maybe not in the best way, maybe you'll want to do
it your own way instead, especially since RC2 is on the way and at the
moment you're using "2.12.0-RC1" as a directory name.  but anyway, but
in any case, here's one possible fix.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -192,10 +192,18 @@ object build extends Build {
     (if (name == "specs2") "" else name) + "> "
   }
 
+  def scalaSourceVersion(scalaBinaryVersion: String) =
+    if (scalaBinaryVersion.startsWith("2.10"))
+      "2.10"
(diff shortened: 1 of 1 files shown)


etorreborre:
That's fine with me. I'll do the cleanup once 2.12 is finally out.

PR #531, opened 2017-01-09
Adds compatibility with Scalaz 7.3.

### LINKED ISSUE #530: Make compatible with Scalaz 7.3
I tried to do this myself by making the 7.1 version work, but it might require a separate set of artifacts?

I ran into two issues. The first (which I fixed) is that things like `Tree.leaf` are gone in Scalaz 7.3 (deprecated in 7.2), but it was easy to replace with `Leaf`. The other (which I couldn’t fix) is that the order of parameters to `tailRecM` changed betwen Scalaz 7.2 and 7.3, so that blows up anything using `BindRec`. I was hoping there was some syntax for tailRecM that would hide the difference in the type class, but no such luck.

I would be happy to submit a PR if anyone thinks they could help me sort out the `tailRecM` issue.

sellout:
Fixes #530.

- use `Leaf` rather than `leaf`, etc. for `scalaz.Tree`
- provide `BindRecʹ` to support both pre- and post-7.3 signatures of `tailrecM`

### DIFF
--- a/common/src/main/scala/org/specs2/control/eff/BindRecʹ.scala
+++ b/common/src/main/scala/org/specs2/control/eff/BindRecʹ.scala
@@ -1,0 +1,14 @@
+package org.specs2.control.eff
+
+import scalaz._
+
+/** Papers over the parameter order change in Scalaz 7.3 */
+trait BindRecʹ[M[_]] extends BindRec[M] {
+
+  def tailrecM[A, B](a: A)(f: A => M[\/[A, B]]): M[B]
+
+  // TODO: Needed for compatibility with Scalaz prior to 7.3
+  def tailrecM[A, B](f: A => M[\/[A, B]])(a: A): M[B] =
+    tailrecM(a)(f)
+
+}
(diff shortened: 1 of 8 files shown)


common/src/main/scala/org/specs2/control/eff/BindRecʹ.scala
```diff
@@ -0,0 +1,14 @@
+package org.specs2.control.eff
+
+import scalaz._
+
+/** Papers over the parameter order change in Scalaz 7.3 */
+trait BindRecʹ[M[_]] extends BindRec[M] {
```
etorreborre: Interesting, I thought that the ability to put `'` in identifiers was only a feature from the Typelevel compiler?

etorreborre:
I don't know why travis is having so many issues running the tests because they pass fine on my laptop. @sellout do you need an official `specs2-scalaz-7.3` version?

sellout:
I don’t think so – we depend on specs2-core and scalaz-specs2, but not specs2-scalaz 😆

sellout:
And thanks so much.

etorreborre:
What I mean is: do you need `specs2-core` with a version `3.8.6-scalaz-7.3`? where `specs2-core` depends on `scalaz-7.3` (at the moment version `3.8.6` for `specs2-core` only depends on `scalaz-7.2.7`).

sellout:
Oh, sorry – I don’t think so. It seems to work fine, with 7.2.7 getting evicted in favor of 7.3.0-M8.

etorreborre:
Ok, thanks.

PR #541, opened 2017-02-05
Array equality fix

noam-almog:
Hi @etorreborre,
I have a suggestion for the array fix you checked in.
On your fix you are forcing diffable by exposing internals by the Diffables companion object.
I prefer to keep the code flow stricter, so in case we have array (which is a special case) we will state it to the compiler and let it resolve the diffable in the same manner it does for all diffables.
WDYT ?

 - Noam

### DIFF
--- a/matcher/src/main/scala/org/specs2/matcher/describe/Diffable.scala
+++ b/matcher/src/main/scala/org/specs2/matcher/describe/Diffable.scala
@@ -67,3 +67,5 @@ trait Diffables {
       }
   }
 }
+
+object Diffables extends Diffables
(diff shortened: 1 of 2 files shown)


etorreborre:
I don't really mind either way but I would leave the `Diffables` object because it could be useful in some client code to invoke a given `Diffable` instance (to solve some type inference issue for example).

noam-almog:
@etorreborre The design I chose (and scala lang) allows the user to specify his own implementations from the outside, but that is not something i think developers will do by themselves, it's not really needed. as for inference problems, i would use this approach only if all things failed. 
anyhow reverted the removal.

PR #542, opened 2017-02-05
reduce duplication

noam-almog:
Glad to see you further integrated the diffable to actually check the result in the EqualityMatcher.
While looking at this code, I want to suggest a different approach, tell me what you think ?

 - Noam


etorreborre:
I think that's a good idea but sealing the traits mean that they cannot be reused outside of specs2.

noam-almog:
@etorreborre removing sealed

etorreborre:
Can you please squash the 2 commits into one with a message like `reduce duplication for comparison results`? Thanks.

noam-almog:
👍

PR #546, opened 2017-02-09
Add a Matcher to assert a Task can fail with a specific Exception

SystemFw:
Since I am not too familiar with the scalaz cross building in specs2, I have only added my changes to the currently active version of TaskMatchers/TaskMatchersSpec. Please let me know if I also need to include it in the other versions.

I have chosen the name `failWith` for my matcher, which is already used in a couple of other places: there should no conflicts, but again, let me know if you'd rather have it named differently.

### DIFF
--- a/scalaz/src/main/scala-scalaz-7.0.x/org/specs2/matcher/TaskMatchers.scala
+++ b/scalaz/src/main/scala-scalaz-7.0.x/org/specs2/matcher/TaskMatchers.scala
@@ -1,9 +1,12 @@
 package org.specs2
 package matcher
 
+import AnyMatchers._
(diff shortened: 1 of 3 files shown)


etorreborre:
`failWith` looks good to me. Can you also please modify the file in the scala-scalaz-7.0.x source directory? I might have to release for `scalaz 7.0.6` as well.

etorreborre:
All good, thanks. Do you need a release soon?

SystemFw:
No worries, this code has been part of my project at work for a while, so I'm in no urgency for a release. I just thought someone else might find it useful, hence the PR.

PR #549, opened 2017-02-15
Fix retry in await helper

jcazevedo:
This PR fixes the retry mechanism in the `await` helper. Since `Future` results are cached, `Future`s weren't being properly retried in case of timeouts.


etorreborre:
Now I understand better the situation and what your fixes is about. Very helpful indeed, thanks!

PR #559, opened 2017-03-15
Add explicit type to implicit val used before definition.

dragos:
Implicits that rely on an inferred return type may not be in scope above their point of definition. In this case, the use comes right before the implicit definition and leads to compilation errors in incremental updates.

You can observe the failure by adding a whitespace to the file and trying to compile (after a full build, so the only file to compile is this one).

See [this thread](https://contributors.scala-lang.org/t/implicits-with-inferred-return-type-lead-to-undefined-behavior/615).

### DIFF
--- a/matcher/src/main/scala/org/specs2/matcher/MatchResultMessages.scala
+++ b/matcher/src/main/scala/org/specs2/matcher/MatchResultMessages.scala
@@ -16,7 +16,7 @@ trait MatchResultMessages {
       case _                  => NeutralMessage(r.message)
     }}
 
-  implicit val MatchResultMessageMonoid = new Monoid[MatchResultMessage] {
+  implicit val MatchResultMessageMonoid: Monoid[MatchResultMessage] = new Monoid[MatchResultMessage] {
     val zero = new EmptySuccessMessage()
     def append(r1: MatchResultMessage, r2: =>MatchResultMessage) = r1 append r2
   }


etorreborre:
Thanks, I suspect it might not be the only one in specs2 since I was not really aware of that when the project started.

dragos:
I think very few people are aware :) This might be the only instance where the return type is missing *and* the implicit is used before the point of definition. There are a lot of implicits with inferred return types, but they should be ok. [scalafix](https://scalacenter.github.io/scalafix/#Rewrites) may help, but it rewrites all of them, not sure if such a big gun is needed here.

dragos:
No idea what the failures mean (2.10 passed, 2.11 and 2.12 failed). Could it be not enough memory?

etorreborre:
Sorry the build is flaky and I don't know yet why 🙈 .

etorreborre:
I also take it that you're experimenting with Hydra and specs2 :-)?

dragos:
Yep. It's in our regression suite and we stepped on this issue. ;-)

dragos:
[Martin's comment on SI-801](https://issues.scala-lang.org/browse/SI-801?focusedCommentId=41419&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41419) and [SI-10229](https://issues.scala-lang.org/browse/SI-10229)

PR #564, opened 2017-04-04
make build Scala 2.13 friendly

SethTisue:
scala-parallel-collections is a module now

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -198,7 +198,8 @@ object build extends Build {
 
   lazy val tests = Project(id = "tests", base = file("tests"),
     settings = moduleSettings("tests") ++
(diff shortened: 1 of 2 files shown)


SethTisue:
these changes are necessary in order to keep specs2 in the Scala 2.13 community build

etorreborre:
Thanks @SethTisue. I don't even remember where and why I'm using parallel collections :-).

SethTisue:
the single usage is in https://github.com/etorreborre/specs2/blob/master/tests/src/test/scala/org/specs2/matcher/TraversableMatchersSpec.scala

etorreborre:
Right, I suspect I must have had a bug with that at some point.

PR #566, opened 2017-04-10
Make TypedEqual object extend the trait

triggerNZ:
I am trying to get the `===` operator available outside of a specification definition. It looks to me like the `TypedEqual` object is for that, but it does not extend the `TypedEqual` trait. That looks like a mistake to me but I'm not overly familiar with the `specs2` codebase so apologies in advance if this is not the case.


etorreborre:
That's a mistake indeed, thanks!

PR #567, opened 2017-04-15
use `Diffable.identical` for `CaseClassPropertyComparison.identical`

tek:
instead of `actual == expected`


etorreborre:
Great, thank you!

etorreborre:
I don't have time right now but I will try to publish a version tomorrow.

PR #568, opened 2017-04-17
use nested `Diffable`s for equivalence in stdlib data classes

tek:
I left the stacktrace stuff untouched, not sure this would be sensible.

### DIFF
--- a/matcher/src/main/scala/org/specs2/matcher/describe/Diffables.scala
+++ b/matcher/src/main/scala/org/specs2/matcher/describe/Diffables.scala
@@ -18,95 +18,104 @@ object PrimitiveDiffable {
 
 }
 
-class EitherDiffable[L : Diffable, R : Diffable](implicit ldi: Diffable[L],
-                                                 rdi: Diffable[R])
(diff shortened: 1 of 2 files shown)


matcher/src/main/scala/org/specs2/matcher/describe/Diffables.scala
```diff
@@ -18,95 +18,104 @@ object PrimitiveDiffable {
 
 }
 
-class EitherDiffable[L : Diffable, R : Diffable](implicit ldi: Diffable[L],
-                                                 rdi: Diffable[R])
+class EitherDiffable[L: Diffable, R :Diffable]
```
etorreborre: Could you please maintain the formatting as `L : Diffable, R : Diffable`?

etorreborre: And also I should be more careful reading PRs, I didn't notice the redundancy in implicit params there.

etorreborre:
That looks good, thanks.

tek:
it actually produced a compile error because I use the implicit as a parameter again :)

PR #600, opened 2017-09-14
removed sbt version specification from travis settings

magnolia-k:
Version of sbt should be specified in project/build.properties

In addition, the version of travis' scala is old(2.12.2), so I modified it to match build.sbt(2.12.3).

It is confirmed that travis of my repository has succeeded this time :)


magnolia-k:
In my repository the successful travis' test has failed in this PR so close this PR

PR #602, opened 2017-09-27
Add missing override to MutableBangExample

nigredo-tori:
The default (immutable) implementation for this `!` overload doesn't add the fragment it produces. This means the fragment can be lost when writing tests in mutable style.

This fix should be also ported to master.


etorreborre:
Oh, thanks for this!

PR #605, opened 2017-10-30
Fix multiple typos

JD557:
I ran a spell checker on the whole project and tried to fix as many typos as possible (I had a lot of false positives, so there might still be some typos in there).

Unfortunately, some typos cannot be changed without performing breaking changes, so I left those alone, namely:

* There's no word "splitted", the correct spelling is "split": (http://www.verbix.com/webverbix/English/split.html);
* It's Levenshtein, not Levenhstein (https://en.wikipedia.org/wiki/Levenshtein_distance).

I also did not change some words that I don't know how to spell (e.g. "uncamel case", "un-camel case", "un-camelcase"?)


etorreborre:
Thanks for those fixes :-).

PR #606, opened 2017-10-30
Avoid unnecessary conversions

JD557:
Avoid unnecessary collection conversions (e.g. `List(1,2,3).toList`).

I made this changes mostly because I had already executed IntelliJ's code analysis to find the typos in #605, so I might as well send some fixes.

This analysis shows a lot of warnings. While most of them are a matter of style/personal preference, there are some warnings there that might make sense to address. If you have the time, I think you should run it and see the results.


etorreborre:
And thanks for that too.

PR #608, opened 2017-11-02
Fix ErrorOrOkOps.toErrorFullMessage

zaneli:
Sorry for my trivial mistake...

ref: #607


etorreborre:
Damn, read the code several times and did not catch it either.

PR #609, opened 2017-11-02
specify charset when writing files

asflierl:
The method comment promises to write the file using the UTF-8 charset, but the file is actually written using the platform default charset (which is not UTF-8 on Windows). This causes HTML reports (that use UTF-8 as hardcoded charset in their header) to contain illegal characters.
The proposed fix looks relatively non-contentious to me.


etorreborre:
That looks good to me as well, thanks!

PR #614, opened 2017-11-24
handle Win file paths like *nix paths

asflierl:
see my comments on https://github.com/etorreborre/specs2/commit/efccac81c4784fb19cc081ff37fd2958f1c2f79d

I've tested my "solution" on Windows. While it works mostly fine, it most probably will not handle all corner cases correctly either. For a correct solution, URIs and file paths must be treated differently. Sadly, even the file system API in the Java stdlib handles some file path cases incorrectly. At the very least, this improves compatibility with Win systems noticeably.

You might want to test if everything still works fine on *nix / Mac.

Hope this helps.

### DIFF
--- a/common/shared/src/main/scala/org/specs2/io/DirectoryPath.scala
+++ b/common/shared/src/main/scala/org/specs2/io/DirectoryPath.scala
@@ -30,10 +30,10 @@ case class DirectoryPath(dirs: Vector[FileName], absolute: Boolean) {
     }
 
   /** @return the path for this file as a / separated string */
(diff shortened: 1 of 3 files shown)


etorreborre:
Everything looks good on my side, let's merge!

etorreborre:
Thanks a lot for this contribution, I'm going to make an interim release and very soon an official `4.0.2`.

etorreborre:
@asflierl here is the release you can use for now `https://oss.sonatype.org/content/repositories/releases/org/specs2/specs2-core_2.12/4.0.2-f59ba9b-20171124130950`

PR #619, opened 2017-12-12
Temporarily downgrade parser combinators

mdedetrich:


etorreborre:
It seems that we still get the same issue in the build, isn't it?

mdedetrich:
@etorreborre The latest push should be working, I fixed it on my end

mdedetrich:
@etorreborre All passed

gourlaysama:
FYI: scala-parser-combinators 1.0.7 has been [released][1] with fixed scala-js support for 2.13.0-M2.
 
And 1.1.0 is coming soon with 2.13.0-M3 support :-)

[1]: https://github.com/scala/scala-parser-combinators/releases/tag/v1.0.7

Issue #591, opened 2017-08-22
scala-native support


xuwei-k:

etorreborre:
I knew it. I knew someone would ask for a scala-native build next :-)

etorreborre:
I am going to close this for now because I realistically don't have the time to implement it.

francisdb:
Tried some things and first blocker is [scalacheck not yet being published for scala native](
https://github.com/rickynils/scalacheck/issues/396)

lolgab:
Scalacheck author didn't reply about publishing scalacheck native on github issue, so I published myself a version on maven under my groupid: "com.github.lolgab" %%% "scalacheck" % "0.4.1"

lolgab:
I published scala-parser-combinators artifact too: "com.github.lolgab" %%% "scala-parser-combinators" % "1.1.2"

PR #620, opened 2017-12-13
Clean up javascript libraryDependencies

mdedetrich:
This PR attempts to clean up the specs2 build a bit. A lot more can be done, but I think this is better done coinciding with https://github.com/etorreborre/specs2/issues/591 (which I will plan to do a bit later)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -115,14 +115,22 @@ lazy val common = crossProject.in(file("common")).
     libraryDependencies ++=
       depends.reflect(scalaOrganization.value, scalaVersion.value) ++
       depends.paradise(scalaVersion.value) ++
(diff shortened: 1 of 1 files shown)


mdedetrich:
@etorreborre This PR is now complete

The idea is that instead of mixing js dependencies and JVM dependencies, JVM only projects (designated as `.jvm`) only depend on JVM sources and the same applies for Javascript.

You run into a lot of weird issues if you mix Scala.js maven jars and Java maven jars, this PR attempts to resolve this.

After this is merged, another PR will be done to migrate to https://github.com/portable-scala/sbt-crossproject which we need for scala-native

etorreborre:
Thanks Matthew.

PR #634, opened 2018-02-14
update and cleanup scala-parser-combinators dependency setting

xuwei-k:


etorreborre:
Thanks. There is one failing test after the upgrade to 2.13.0-M2, I am investigating it.

PR #643, opened 2018-03-25
Fix for failing JUnitXmlPrinterSpec 

hmemcpy:
This PR reverts a (possibly unintentional) removal of `content` assignment  to the mock filesystem, that was part of 389ed91317d6cbb98a081cb9fd165cd26f42e982 (upgrad to scala 2.13.0-M3).


hmemcpy:
Yeah, I would imagine this produces a warning, since the assignment return type is Unit. But, it works :)

PR #645, opened 2018-03-25
Fixing a compilation error with Scala 2.13.x

hmemcpy:
This fixes an error that occurs in Scala 2.13.0-M3, it fails to find the `out` field in scope, resulting in the following error:
```
[error] .../JUnitXmlPrinterSpec.scala:111: unknown parameter name: out
[error]         Operations.ok(out = content)
```

This PR fixes it by explicitly specifying `this.` keyword. 

This is also a possible bug/regression in Scala 2.13.


hmemcpy:
I got curious whether it's a Scala bug or not, so I asked on gitter, and got pointed to this issue: https://github.com/scala/scala/pull/6089

It's not a bug, the opposite. It's deprecated behavior!

etorreborre:
Thanks!

PR #647, opened 2018-04-06
avoid procedure syntax

xuwei-k:


etorreborre:
Thanks, did you use a warning/plugin to detect all of those?

PR #657, opened 2018-05-14
fix build (discovered in Scala community build)

SethTisue:
this reverts a piece of ebc48e9276 that I don't even understand
why it compiled. the type of `ghreleaseNotes` is `String => String`

in the context of the community build, what happens is that dbuild
manipulates the `version` setting to something long and gruesome like
4.2.0-DBUILDXD306556371FB43C35AA0C7F2F61228127D1882C3


SethTisue:
you should probably test it with sbt-github-release though before merging, since I'm not familiar with that plugin.

etorreborre:
I did that change kind of blindly because sbt was not happy about the previous version. I'll integrate your change and try to fix the warning properly.

PR #664, opened 2018-06-09
dependency version bumps

SethTisue:
these are the versions that are published for 2.13.0-M4, so this is
needed to prepare for M4 support (once ScalaCheck has published)


etorreborre:
Thanks @SethTisue for warning me when ScalaCheck is ready in case I miss it.

SethTisue:
ScalaCheck has now been published for M4

PR #667, opened 2018-06-11
update scalaz 7.2.24

xuwei-k:
prepare Scala 2.13.0-M4


etorreborre:
Thanks, I have incorporated all of your changes + mines and now master should be ok for the latest 2.13.

etorreborre:
I am trying to publish now, hopefully it would go better than yesterday.

PR #668, opened 2018-06-11
fix scalacOptions for Scala 2.13.0-M4

xuwei-k:
some options removed since Scala 2.13.0-M4
- https://github.com/scala/scala/commit/ad25805de5c63084348348aabefdf184927806c2
- https://github.com/scala/scala/commit/7d5e0b01c7f645e4f727f704a18f93ce5c69a9dd


etorreborre:
Thank you I'm just trying to publish for 2.13.0-M4 right now :-)

PR #675, opened 2018-06-21
Cats NonEmptyList Sized

keirlawson:


cats/shared/src/main/scala/org/specs2/matcher/NonEmptyListMatchers.scala
```diff
@@ -0,0 +1,14 @@
...
+import org.specs2.data.Sized
+
+trait NonEmptyListMatchers {
+
+  implicit def nonEmptyListIsSized[T]: Sized[NonEmptyList[T]] = new Sized[NonEmptyList[T]] {
+    override def size(t: NonEmptyList[T]): Int = t.length
+  }
+}
+
+object NonEmptyListMatchers {
```
etorreborre: It seems that this can be removed.

cats/jvm/src/test/scala/org/specs2/matcher/NonEmptyListMatchersSpec.scala
```diff
@@ -0,0 +1,14 @@
+package org.specs2
+package matcher
+
+import cats.data.NonEmptyList
+
+class NonEmptyListMatchersSpec extends Spec with NonEmptyListMatchers { def is =
+  s2"""
```
etorreborre: Just a formatting thing, can you please align the `s2`: `def is = s2""" ... `
and add a newline at the end of the file?

keirlawson:
amended as requested

PR #678, opened 2018-06-22
Added matchers for cats.effect.IO

GW00D:
Added matchers for cats.effect.IO

### DIFF
--- a/cats/jvm/src/test/scala/org/specs2/matcher/IOMatchersSpec.scala
+++ b/cats/jvm/src/test/scala/org/specs2/matcher/IOMatchersSpec.scala
@@ -1,0 +1,29 @@
+package org.specs2
+package matcher
+
+import cats.effect.IO
+import org.specs2.specification.Retries
+
+import scala.concurrent.duration._
+
+class IOMatchersSpec extends mutable.Specification with IOMatchers with ResultMatchers with Retries {
+  sequential
+
+  "It is possible to check the execution of tasks" >> {
+    "check the return value" >> {
+      { IO(1) must returnValue(1) }
+      { (IO(1) must returnValue(2)) returns "1 != 2" }
+    }
+
+    "check that the task finishes before a given time" >> {
+      { IO(1) must returnBefore(10.millis) }
+      { (IO { Thread.sleep(50); 1 } must returnBefore(10.millis)) returns "Timeout after 10 milliseconds" }
+    }
+
+    "check the return value with a timeout" >> {
+      { IO { Thread.sleep(5); 1 } must returnBefore(10.millis).withValue(1) }
+      { IO { Thread.sleep(5); 1 } must returnValue(1).before(10.millis) }
+    }
+  }
+}
+
--- a/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
+++ b/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
@@ -1,0 +1,67 @@
+package org.specs2.matcher
+
+import cats.effect.IO
+import org.specs2.matcher.ValueChecks.valueIsTypedValueCheck
+import org.specs2.matcher.describe.Diffable
+
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.FiniteDuration
+import org.specs2.text.NotNullStrings._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+trait IOMatchers {
+  def returnOk[T]: IOMatcher[T] =
+    attemptRun(ValueCheck.alwaysOk, None)
+
+  def returnValue[T](check: ValueCheck[T]): IOMatcher[T] =
+    attemptRun(check, None)
+
+  def returnBefore[T](duration: FiniteDuration): IOMatcher[T] =
+    attemptRun(ValueCheck.alwaysOk, Some(duration))
+
+  private[specs2] def attemptRun[T](check: ValueCheck[T], duration: Option[FiniteDuration]): IOMatcher[T] =
+    IOMatcher(check, duration)
+
+  case class IOMatcher[T](check: ValueCheck[T], duration: Option[FiniteDuration]) extends Matcher[IO[T]] {
+    def apply[S <: IO[T]](e: Expectable[S]) = duration.fold(checkIO(e))(checkIOWithDuration(e,_))
+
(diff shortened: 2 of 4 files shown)


cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import org.specs2.text.NotNullStrings._
+import scala.concurrent.ExecutionContext.Implicits.global
+import cats.effect.Timer
+
+trait IOMatchers {
+
+  implicit val timer = Timer[IO]
```
etorreborre: Can you please add the return type?
```
implicit val timer: Timer[IO] = 
  Timer[IO]
```
This is recommended practice for implicit values (I think it is now mandatory in Dotty)

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  implicit def returnOk[T]: IOMatcher[T] =
```
etorreborre: Why do those matchers need to be implicit?

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,73 @@
...
+    attemptRun(ValueCheck.alwaysOk, None)
+
+  implicit def returnValue[T](check: ValueCheck[T]): IOMatcher[T] =
+    attemptRun(check, None)
+
+  implicit def returnBefore[T](duration: Duration): IOMatcher[T] =
```
etorreborre: I think we should better require a `FiniteDuration` here.

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    private def errorResult[S <: IO[T]](e: Expectable[S])(t: Throwable): MatchResult[S] = {
+      val message = "an exception was thrown "+t.getMessage.notNull+" "+t.getClass.getName
+      result(false, message, message, e)
+    }
+
+    private def checkResult[S <: IO[T]](e: Expectable[S])(t: T): MatchResult[S] =
+      result(check.check(t), e)
+
+    def asFiniteDuration(d: Duration): FiniteDuration = Duration.fromNanos(d.toNanos)
```
etorreborre: Then we won't need that line.

guide/src/test/scala/org/specs2/guide/matchers/IOMatchers.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object IOMatchers extends UserGuideCard {
+  def title = "IO Matchers"
+  def text = s2"""
+  There are some basic matchers to check `cats.effect.IO` instances:
+
+ * `returnOk` to check if an `IO` completes successfully
+ * `returnBefore(duration)` to check if an `IO` completes successfully before the specified duration
+ * `returnValue(matcher)` to check if an `IO` successfully returns a value that satisfies the matcher
+ * `returnValue(function: A => AsResult[B])` to check if an `IO` completes successfully with a value `a`, where `function(a)` returns a successful [Result](org.specs.guide.AsResultTypeclass.html). This can be used to do conditional assertions that depend on which value is returned by your `IO`
```
etorreborre: Is it possible to break this line to fit the screen?

cats/jvm/src/test/scala/org/specs2/matcher/IOMatchersSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+      { (IO(1) must returnValue(2)) returns "1 != 2" }
+    }
+
+    "check that the task finishes before a given time" >> {
+      { IO(1) must returnBefore(10.millis) }
+      { (IO { Thread.sleep(50); 1 } must returnBefore(10.millis)) returns "Timeout after 10 milliseconds" }
+    }
+
+    "check the return value with a timeout" >> {
+      { IO { Thread.sleep(5); 1 } must returnBefore(10.millis).withValue(1) }
```
etorreborre: Nice!

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
GW00D: Thanks :+1:

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.FiniteDuration
+import org.specs2.text.NotNullStrings._
+import scala.concurrent.ExecutionContext.Implicits.global
+import cats.effect.Timer
+
+trait IOMatchers {
+
+  val timer: Timer[IO] = Timer[IO]
```
etorreborre: I don't see where the `timer` is used, is it needed for anything?

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def returnOk[T]: IOMatcher[T] =
+    attemptRun(ValueCheck.alwaysOk, None)
+
+  implicit def returnValue[T](check: ValueCheck[T]): IOMatcher[T] =
```
etorreborre: Is the `implicit` necessary here?

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,71 @@
...
+    attemptRun(check, None)
+
+  implicit def returnBefore[T](duration: FiniteDuration): IOMatcher[T] =
```
etorreborre: And there?

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
GW00D: Sorry - thanks for catching that!

GW00D:
@etorreborre Updated again

etorreborre:
Great thank you!

Issue #633, opened 2018-02-09
Regression 4.0.2: Stack Overflow


peter-empen:
Do you have a suggestion how to resolve the following issue that occurs on switching from from 3.8. to 4.0.2 and working extensively with the finder:

```
java.lang.StackOverflowError
	at org.specs2.fp.Traverse$$anon$1.$anonfun$traverseImpl$3(Traverse.scala:38)
	at org.specs2.fp.Monad.ap(Monad.scala:19)
	at org.specs2.fp.Monad.ap$(Monad.scala:18)
	at org.specs2.control.eff.EffImplicits$$anon$1.ap(Eff.scala:141)
	at org.specs2.fp.Applicative.ap2(Applicative.scala:18)
	at org.specs2.fp.Applicative.ap2$(Applicative.scala:17)
	at org.specs2.control.eff.EffImplicits$$anon$1.ap2(Eff.scala:141)
	at org.specs2.fp.Applicative.apply2(Applicative.scala:33)
	at org.specs2.fp.Applicative.apply2$(Applicative.scala:32)
	at org.specs2.control.eff.EffImplicits$$anon$1.apply2(Eff.scala:141)
	at org.specs2.fp.Traverse$$anon$1.traverseImpl(Traverse.scala:38)
	at org.specs2.fp.Traverse$$anon$1.$anonfun$traverseImpl$3(Traverse.scala:38)
```

etorreborre:
Ouch, the first idea would be to run with `-Xss4m` in order to get a bigger stack.

peter-empen:
Certainly. But, provided you can exclude an infinite loop, what about moving to a tail recursive implementation?

etorreborre:
I suspect that the part where there's a loop is a part that I don't know how to make tail-recursive. I will try to at least put a work-around.

peter-empen:
Yeh, that is often too hard a job. I sometimes also spent a lot of time to get specific algorithms tail-recursive.
Just for curiosity, could you point me to that method?

etorreborre:
Sure, but just for curiosity then :-). Specs2 is running an old version of the eff library internally and its `ap` method for the `Eff` applicative is: https://github.com/etorreborre/specs2/blob/master/common/shared/src/main/scala/org/specs2/control/eff/Eff.scala#L177.

[...]

PR #683, opened 2018-07-03
Take constant stack while running examples in parallel (fix #682)

### LINKED ISSUE #682: Stack overflow in `TimedFutureApplicative.ap`
A test suite with a lot of examples occasionally fails with this trace:
https://gist.github.com/edmundnoble/69d8805baf25deebabdfe684d73b2554

I believe this is because the `Applicative` instance for `TimedFuture` does not properly bounce on the stack between examples, meaning that specs2 consumes stack space linear in the number of examples.

I have a fix incoming.

If this is actually because of an infinite loop, my fix won't fix anything. Unfortunately, I can't reliably reproduce the failure, and I can't discern anything more accurate from the trace anyway.

edmundnoble:
This should fix #682 by suspending the `runNow` calls to the execution context, so that constant stack is taken. ~~I am going to check today that this fixes the issue, but I can't see how it wouldn't.~~

I can't check that this fixes my issue directly, because it's nondeterministic.

### DIFF
--- a/common/shared/src/main/scala/org/specs2/control/eff/FutureEffect.scala
+++ b/common/shared/src/main/scala/org/specs2/control/eff/FutureEffect.scala
@@ -42,8 +42,8 @@ object TimedFuture {
 
     def ap[A, B](fa: =>TimedFuture[A])(ff: =>TimedFuture[(A) => B]): TimedFuture[B] = {
       val newCallback = { es: ExecutorServices =>
-        val ffRan = ff.runNow(es)
-        val faRan = fa.runNow(es)
+        val ffRan = Future(ff.runNow(es))(es.executionContext).flatten
+        val faRan = Future(fa.runNow(es))(es.executionContext).flatten
         faRan.flatMap(a => ffRan.map(f => f(a))(es.executionContext))(es.executionContext)
       }
       TimedFuture(newCallback)


edmundnoble:
Actually, I'm going to add a test suite which adds a *bunch* of examples.

edmundnoble:
I can't even reproduce this reliably with an artificial example. I have no idea why or when this codepath in specs2 is being hit; all I know is that we're hitting it, and it isn't stack-safe.

etorreborre:
I am wondering if this is the same cause as #633. My thought was that the Eff.ap implementation is correct. I can still merge your change and publish an interim version to see if that helps.

djspiewak:
Any chance of getting a quick backport-cherrypick-release to the 4.1.x line? We're still stuck on scalacheck 1.13, so we can't upgrade to 4.2.x yet.

etorreborre:
Definitely I'll try to do that tomorrow.

edmundnoble:
Just tested, this fixes the issue. :tada:

etorreborre:
@djspiewak, @edmundnoble please try `4.1.1`.

djspiewak:
@etorreborre I know I've said it before over the years, but I don't say it enough: Once again, you have shown why it is you're such an incredible OSS maintainer, and why I aspire to be just like you when I grow up. 😃

etorreborre:
Thanks @djspiewak :-)

djspiewak:
@etorreborre I can't seem to resolve `4.1.1`? Has it just not synced yet?

djspiewak:
It also appears that there was no tag pushed for 4.1.1, so I can't publish locally either. I'm guessing this is just being delayed by the release issues you mentioned on Twitter. :-)

etorreborre:
@djspiewak actually no I only had some issues with 2.12. That is fixed now and 

 - 2.11 is [here](https://oss.sonatype.org/content/repositories/releases/org/specs2/specs2-core_2.11/4.1.1)
 - 2.12 is [here](https://oss.sonatype.org/content/repositories/releases/org/specs2/specs2-core_2.12/4.1.1)
 - 2.13.0-M3 is [here](https://oss.sonatype.org/content/repositories/releases/org/specs2/specs2-core_2.13.0-M3/4.1.1)

Also publishing today was a lot faster, I don't know if that's because I'm doing from home or if because it's Saturday but it was a lot more bearable today.

djspiewak:
@etorreborre Looks like it is still broken: https://gist.github.com/492979f382430444a76bc38d743f5a89 Got that with specs2 4.1.1.

PR #686, opened 2018-07-09
Use constant stack running TimedFuture sequentially, too

edmundnoble:
I was looking at the wrong code in the stack trace; I saw that it was `ap`, but I didn't know it was a derived `ap` (and there are no frames in `bind`, which is being used to implement it). As well, I didn't understand the circumstances required to reproduce the bug on the branch which reproduced it (it required running the same test suite twice, oddly, but doing so reproduced it 10/10 times). Now that I've reproduced the bug ~10 times, then made this patch and failed to reproduce it ~10 times, I believe we have a winner. Sorry about that. Though, that previous patch fixed a legitimate stack issue too.

A backport to 4.1 would be excellent this time as well.


etorreborre:
Great thanks for the new patch, I will do both releases tomorrow.

Issue #691, opened 2018-07-23
JUnitRunner: beforeAll and afterAll steps registered as a test in junit


kaatzee:
When there is implemented beforeAll or/and afterAll method in test they are registered as test called 'step' in junit. 

I have provided project (https://github.com/kaatzee/specs2Issues) to reproduce this issue. Run https://github.com/kaatzee/specs2Issues/blob/master/src/test/scala/s2i/beforeandafterallastest/HelloWorldSpec.scala test with JUnitRunner.

etorreborre:
fixed with #700

PR #692, opened 2018-07-23
beforeAll and afterAll don't emit step test anymore in JUnitRunner

kaatzee:
Fix proposal for #691

### DIFF
--- a/junit/jvm/src/test/scala/org/specs2/runner/JUnitRunnerSpec.scala
+++ b/junit/jvm/src/test/scala/org/specs2/runner/JUnitRunnerSpec.scala
@@ -1,53 +1,84 @@
 package org.specs2
 package runner
 
-import org.junit.runner.{Description, Result, RunWith}
 import org.junit.runner.notification._
-import main.Arguments
-import specification.core.{Env, OwnEnv}
-import control._
-import ExecuteActions._
+import org.junit.runner.{Description, Result, RunWith}
+import org.specs2.control.ExecuteActions._
+import org.specs2.main.Arguments
+import org.specs2.specification.BeforeAfterAll
+import org.specs2.specification.core.{Env, OwnEnv}
 
-class JUnitRunnerSpec(val env: Env) extends Specification with OwnEnv { def is = s2"""
+import scala.collection.mutable.ListBuffer
 
- The Junit runner must run all linked specifications if 'all' is set on the command line $allSpecifications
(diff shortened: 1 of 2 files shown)


junit/jvm/src/test/scala/org/specs2/runner/JUnitRunnerSpec.scala
```diff
@@ -58,3 +89,17 @@ object LinkedJUnitSpec extends mutable.Specification {
...
+@RunWith(classOf[JUnitRunner])
+class JUnitWithBeforeAfterAllSpecification extends Specification with BeforeAfterAll {
+  def is =
+    s2"""
+
+  one example $ok
+
+"""
+
+  override def beforeAll(): Unit = {}
```
etorreborre: I would rather write 
```scala
def beforeAll: Unit = ()
```
here to stay consistent with other places in the code.

etorreborre:
This looks like the right thing to do indeed, thanks.

etorreborre:
The only issue is: how do we report a step error? We should probably do a savage `println` in that case.

kaatzee:
According to reporting error in step. I have looked to JUnit code([ParentRunner.java](https://github.com/junit-team/junit4/blob/02c328028b4d32c15bbf0becc9838e54ecbafcbf/src/main/java/org/junit/runners/ParentRunner.java#L390), [EachTestNotifier.java](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java#L19)) and also enhanced a little bit my example([RunListenerTest.java](https://github.com/kaatzee/specs2Issues/blob/master/src/test/java/s2i/junitlistener/RunListenerTest.java), [JUnitExecutionListener.java](https://github.com/kaatzee/specs2Issues/blob/master/src/main/java/s2i/junitlistener/JUnitExecutionListener.java)) and it looks like that for exception in beforeAll just calling method `fireTestFailure` with proper description name is enough. In such case description name should be set to `"RunListenerTest.java"` instead of `"step(RunListenerTest.java)"`. Here ([JUnit4TestListener.java](https://github.com/JetBrains/intellij-community/blob/master/plugins/junit_rt/src/com/intellij/junit4/JUnit4TestListener.java#L212) we can see that for example Intellij is prepared for such scenario.
I can try to prepare some fix proposal for this but only after weekend.
Have a nice weekend!

kaatzee:
Replaced by #700.

Issue #693, opened 2018-07-23
JUnitRunner: test marked as pendingUntilFixed not displayed correctly in junit.


kaatzee:
Issue is reproduced here: https://github.com/kaatzee/specs2Issues/blob/master/src/test/scala/s2i/pendinguntilfixed/HelloWorldSpec.scala. Just run it with JUnitRunner.

According to https://memorynotfound.com/add-junit-listener-example/ ignored test should be registered in JUnit without firing 'start'. This code is also provided in https://github.com/kaatzee/specs2Issues/blob/master/src/test/java/s2i/junitlistener/RunListenerTest.java.

etorreborre:
fixed with #700

PR #694, opened 2018-07-23
junit test ignored registered properly

kaatzee:
This is fix proposal for #693

This commit is built on top of: #692


etorreborre:
That looks good. I must say I couldn't find a precise spec for the JUnit notifier API. I will rebase this PR on the previous one.

PR #696, opened 2018-07-29
better levels for non mutable specifications

kaatzee:
Non mutable specifications run by JUnitRunner were not properly printed. Here is example: https://github.com/kaatzee/specs2Issues/blob/master/src/test/scala/s2i/nonmutablespecification/NonMutableIssueSpec.scala (just run it via JUnitRunner).

I have corected calculating 'levels' of 'fragments' and now it looks much better. Not sure if it is desired output, but it is similar to outputs from other notifiers or printers.

Also in Levels trait there is 'fold' method which had had the same (almost) behaviour, but I haven't changed it. It is used by HtmlPrinter, but I haven't checked if it suffers for the same issue (probably yes?).

### DIFF
--- a/core/jvm/src/test/scala/org/specs2/specification/LevelsSpec.scala
+++ b/core/jvm/src/test/scala/org/specs2/specification/LevelsSpec.scala
@@ -34,7 +34,7 @@ class LevelsSpec(ee: ExecutionEnv) extends Spec { def is = s2"""
       "|",
       "`- Fragment(e2)")
 
-    def e1: MatchResult[Option[Tree[Fragment]]] = treeMap(new mutableSpec { "t1" >> { "e1" in ok; "e2" in ok } }.is.fragments)(mapper)(ee) must beDrawnAs(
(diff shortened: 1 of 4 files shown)


etorreborre:
As far as I know the HTML output is correct, since you must have this branch opened somewhere can you please try to run your `NonMutableIssueSpec` with the `html` argument and check the output to see how the output is affected? Thanks

kaatzee:
I have added two tests for HtmlBodyPrinter and it looks like that HTML output is not affected. But... Levels in HtmlPrinter are almost not used at all. It is especially visible in test for mutable spec where I would expect several `<ul></ul>` tags but there are none. I am not sure if these levels are needed here.

I have also changed way of getting levels for fragments. Now it looks simpler. Tests had been added before and it looks like it behaves the same way.

etorreborre:
I think that in the case of a mutable spec a `Text` fragment is not followed directly by an example so that one doesn't get indented. In any case I tested your changes on a few examples and in particular on the user guide which seems ok.

I wonder if we could remove the blank lines between examples in the html report for a mutable spec
image
but at least we have the same behaviour on the current master. So I'll say let's merge! Thanks.

kaatzee:
About blank lines, I have played a little with and have come with some proposal: #704. Take a look and decide if it is what you have on mind. 

Also thanks for checking!

PR #700, opened 2018-08-13
Error in before all properly reported.

kaatzee:
This is fix proposal for both #691 and #693. It is replacement for my previous pull requests (#692, #694)


etorreborre:
This looks good to me, thanks!

PR #704, opened 2018-08-19
in HtmlPrinter delete line between examples

kaatzee:
In html output there is a line between examples:

This PR deletes this line.

### DIFF
--- a/html/src/main/scala/org/specs2/reporter/HtmlBodyPrinter.scala
+++ b/html/src/main/scala/org/specs2/reporter/HtmlBodyPrinter.scala
@@ -14,6 +14,7 @@ import form._
 import control._
 import ExecuteActions._
 import org.specs2.concurrent.ExecutionEnv
+import org.specs2.control.producer.{Transducer, Transducers}
 import org.specs2.text.AnsiColors
 import origami._
 import org.specs2.time.SimpleTimer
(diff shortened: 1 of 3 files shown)


html/src/main/scala/org/specs2/reporter/HtmlBodyPrinter.scala
```diff
@@ -30,14 +31,22 @@ trait HtmlBodyPrinter {
     val title = spec.name
     type HtmlState = (String, Level)
 
+    val deleteLineBetweenExamples: Transducer[ActionStack, Fragment, Fragment] = producer =>
```
etorreborre: I think it is worth adding some comments explaining that:
 - when the specification is built some `br` fragments are inserted between examples so that they are
 rendered properly in the console
 - but we rendering in HTML we already use `<ul/>` to display examples so those `br` become unnecessary.

html/src/main/scala/org/specs2/reporter/HtmlBodyPrinter.scala
```diff
@@ -30,14 +31,22 @@ trait HtmlBodyPrinter {
...
+      producer.pipe(Transducers.zipWithPreviousAndNext).filter {
```
etorreborre: `zipWithPreviousAndNext` is quite a useful operator for streaming isn't it :-)?

kaatzee: Yes, it is :) I'd tried to implement this other way until I noticed this operator and then it was just easier :)

html/src/main/scala/org/specs2/reporter/HtmlBodyPrinter.scala
kaatzee: I've added some comment.

PR #705, opened 2018-08-23
Prevent double-reporting of junit test run status 

### LINKED ISSUE #440: Failing to integrate with Bazel build tool
I've been playing with the [Bazel](https://github.com/bazelbuild/bazel) build tool, and have a running example compiling and testing Scala code [here](https://github.com/orrsella/bazel-example). One problem I'm still facing is that for some reason specs2 tests that should fail aren't failing the build. I do see them as executed (by `println`-ing from the test and seeing it in test logs), but the failing test doesn't seem to fail the build. The curious thing is that ScalaTest tests do fail the build correctly. The tests are run by invoking `JUnitCore` with a list of classes or a single tests that is `@RunWith` a custom `Suite`.

After trying to debug it, I have a suspicion that maybe the `JUnitRunner` is doing something "fishy". I created a simple project with a few tests, and then ran it manually from the command line:

```scala
package com.example

import org.junit.runner.RunWith
import org.specs2.mutable.Specification
import org.specs2.runner.JUnitRunner

@RunWith(classOf[JUnitRunner])
class Specs2Test extends Specification {
  "Foo" should {
    "bar" in {
      1 === 1
    }
  }
}
```

```bash
$ java -cp .:lib/* org.junit.runner.JUnitCore com.example.Specs2Test
JUnit version 4.12
.
Time: 0

OK (0 tests)


Time: 0.338

OK (1 test)
```

Notice the seemingly double execution of tests. It's also present for a failing test:

```bash
$ java -cp .:lib/* org.junit.runner.JUnitCore com.example.Specs2FailingTest
JUnit version 4.12
.E
Time: 0

OK (0 tests)


Time: 0.399
There was 1 failure:
1) Foo should::bar(com.example.Specs2Test)
java.lang.Exception: '1' is not equal to '2'
	at com.example.Specs2Test$$anonfun$1$$anonfun$apply$3.apply(Specs2Test.scala:12)
	at com.example.Specs2Test$$anonfun$1$$anonfun$apply$3.apply(Specs2Test.scala:12)

FAILURES!!!
Tests run: 1,  Failures: 1
```

For reference, here's the output from ScalaTest (notice only 1 tests block):

```bash
$ java -cp .:lib/* org.junit.runner.JUnitCore com.example.ScalaTest
JUnit version 4.12
.
Time: 0.15

OK (1 test)
```

A failing ScalaTest test also has only 1 tests block. Both specs2 and ScalaTest return exit code of `1` when the tests are set to fail.

I'm sort of out of ideas. Would appreciate any thoughts on this. Thanks!

hmemcpy:
Fixes #440 again, seems it was broken sometime ago.
This fix makes sure to evaluate the boolean flag in the main thread, so the stack trace of the actual calling process is available.

### DIFF
--- a/common/shared/src/main/scala/org/specs2/control/ExecutionOrigin.scala
+++ b/common/shared/src/main/scala/org/specs2/control/ExecutionOrigin.scala
@@ -21,6 +21,10 @@ trait ExecutionOrigin extends Stacktraces {
   lazy val isExecutedFromJUnitCore = isExecutedFrom("org.junit.runner")
   /** return true if the current test is executed with Scala.js */
   lazy val isExecutedFromScalaJs = isExecutedFrom("org.scalajs")
+  /** return true if the current test is executed with Bazel */
+  lazy val isExecutedFromBazel = isExecutedFrom("com.google.testing.junit")
+
+  lazy val excludeFromReporting: Boolean = isExecutedFromJUnitCore || isExecutedFromBazel
 
   /** try to approximate if a specification is a specs2 one or scalaz one by passing name = org.specs2 or name = scalaz */
   def isSpecificationFromSpecs2orScalaz(st: Seq[StackTraceElement]) = {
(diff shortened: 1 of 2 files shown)


etorreborre:
All good thank you. How fast do you need a release?

hmemcpy:
The sooner the better :))
We have some folks blocked by it.

etorreborre:
Ok `4.3.4` is published now (I'm at home and it also takes forever to publish).

hmemcpy:
Amazing, thank so much for the quick release!

PR #707, opened 2018-09-13
Refactor IOMatchers to add Generic

diesalbla:
### Description of changes

We split the IOMatchers trait to make it independent from IO.

- We extract a generic `RuntimedMatchers` trait, which is generic on  the effect type `F[_]`, and has most methods from `IOMatchers`.
- The generic `RuntimedMatchers` defines two methods, that act as placeholders for `IO.timeout` and `IO.unsafeRunSync`.
- To the `RunTimedMatchers` trait, we add an inner `TimedMatcher` class, that has most of the logic from the `IOMatcher` class. This class has methods, akin to `checkIOWithDuration` and `checkIO`, but with different names.

Having moved most of the logic to `RuntimedMatchers`, the `IOMatcher` becomes just a specific instance of `RunTimedMatchers[IO]`, and the two methods just call the IO methods.

- We keep an `IOMatcher` inner class inside `IOMatchers`, for   compatibility. This class keeps the two `checkIOWithDuration` and `checkIO` methods above.

### Related tickets

- This changes were first proposed in a [Pull Request to `http4s`](https://github.com/http4s/http4s/pull/2080), which contains a copy of `IOMatchers`. 
- A proposal to generalise some methods of `cats.effect.IO` was made and discussed in [this issue](https://github.com/typelevel/cats-effect/issues/230). Nevertheless, since no appropriate type-class at present, declaring abstract methods in `RuntimedMatchers` is the next best.

### DIFF
--- a/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
+++ b/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
@@ -7,61 +7,82 @@ import org.specs2.matcher.describe.Diffable
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration.FiniteDuration
 import org.specs2.text.NotNullStrings._
-import scala.concurrent.ExecutionContext.Implicits.global
 
-trait IOMatchers {
-  def returnOk[T]: IOMatcher[T] =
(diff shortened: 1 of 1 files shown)


etorreborre:
This makes sense, thanks.

PR #710, opened 2018-10-01
allow building on both JDK 8 & JDK 11

SethTisue:
for me this is coming up in the context of trying to run the Scala 2.12 community build on JDK 11 (as well as 8, as usual)

but it's also worth having in general, I think, because your users and contributors will increasingly be on 11. (but also many will be on 8 for a while yet to come, since for now it remains the most recommended, best supported JDK for doing Scala work.)


etorreborre:
Thanks Seth.

PR #711, opened 2018-10-02
upgrade cats + cats-effect to versions community build expects

SethTisue:
this should be reviewed by somebody who actually knows what they're
doing. I understand this about as well as your dog understands CNN

I'm just trying to get the community build green

### DIFF
--- a/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
+++ b/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
@@ -71,6 +71,10 @@ trait RunTimedMatchers[F[_]] {
 trait IOMatchers extends RunTimedMatchers[IO] {
 
   import scala.concurrent.ExecutionContext.Implicits.global
+  implicit val catsEffectTimer: cats.effect.Timer[IO] =
+    IO.timer(global)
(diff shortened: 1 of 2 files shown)


SethTisue:
maybe the cats dependency should be 1.4.0 instead of 1.3.1, idk

cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -71,6 +71,8 @@ trait RunTimedMatchers[F[_]] {
 trait IOMatchers extends RunTimedMatchers[IO] {
 
   import scala.concurrent.ExecutionContext.Implicits.global
+  implicit val timer: cats.effect.Timer[IO] = IO.timer(global)
```
etorreborre: I think this is ok but could you rename the variables as `catsEffectTimer` and `catsEffectContextShift` please? I want to avoid name clashes when that trait is imported into a specification, especially with the `timer` name which might be already used by something else.

SethTisue: done

etorreborre:
Thank you.

SethTisue:
scala/community-builds@d1e2e0dbc690dd50aadd957bc7e85d007689561b

PR #724, opened 2018-12-05
unpack match results in cats-effect IO

bpholt:
This enables tests using syntax like this:

```scala
"an IO[MatchResult[_]] is implicitly a Result" >> {
  for {
    x <- IO.pure(42)
  } yield {
    x must_== 42
  }
}
```


bpholt:
@etorreborre That makes sense. I was able to add this function easily enough, and it works for both JS and JVM:

```scala
implicit def ioAsExecution[R : AsResult]: AsExecution[IO[R]] = new AsExecution[IO[R]] {
  def execute(r: =>IO[R]): Execution = Execution.withEnvAsync(_ => r.unsafeToFuture())
}
```

Once I got the tests in the `specs2-cats` running for both JVM and JS, though, I realized that the existing `IO` matchers currently run synchronously:

https://github.com/etorreborre/specs2/blob/2993e8bd9f29d4115cff9f6207e551c19d2dc7e0/cats/shared/src/main/scala/org/specs2/matcher/IOMatchers.scala#L79-L80

which means only the untimed matchers work on JS (and I think only when the `IO` is actually pure and wouldn't block).

I made some changes that remove the `returnBefore` method from `IOMatchers` on the JS platform. It wasn't obvious to me how to create a `Matcher[IO[T]]` (via `TimedMatcher` or otherwise) that stayed async and therefore would work on JS. If anyone has ideas on that, I'm open to them; otherwise, it's possible that the JS `IOMatchers` can be further simplified to clean up some of the remnants of `TimedMatcher` etc. (I tried to break the changes up into logical chunks in my commits, if it helps to consider these points separately.)

cats/js/src/main/scala/org/specs2/matcher/IOMatchers.scala
```diff
@@ -0,0 +1,71 @@
...
+
+trait IOMatchers extends RunFMatchers[IO] {
+  implicit val ee: ExecutionEnv
+  implicit def catsEffectTimer(implicit ee: ExecutionEnv): cats.effect.Timer[IO] =
+    IO.timer(ee.ec)
+  implicit def catsEffectContextShift(implicit ee: ExecutionEnv): cats.effect.ContextShift[IO] =
+    IO.contextShift(ee.ec)
+
+  implicit def ioAsExecution[R: AsResult]: AsExecution[IO[R]] = new AsExecution[IO[R]] {
+    def execute(r: => IO[R]): Execution = Execution.withEnvAsync(_ => r.unsafeToFuture())
```
etorreborre: I wonder if should shift the `IO` computation here to the execution context passed in `withEnvAsync` (which is otherwise ignored with `_`) to make sure that the future that is created here runs with the right execution context

common/js/src/main/scala/org/specs2/reflect/Classes.scala
```diff
@@ -12,8 +13,16 @@ trait Classes extends ClassOperations {
     throw new Exception("Classes.createInstance: no js implementation")
 
   def createInstance[T <: AnyRef](className: String, loader: ClassLoader, defaultInstances: =>List[AnyRef] = Nil)(implicit m: ClassTag[T]): Operation[T] = {
+    def newInstanceWithOptionalExecutionEnvironment() =
```
etorreborre: Could you please extract this method outside `createInstance` and give it a signature `def newInstanceWithOptionalExecutionEnvironment: T` (I generally don't use `()` in specs2)?

etorreborre:
@bpholt I restarted Travis and the build is green now. I am on vacations until the end of the week so I will try to have a deeper look at your questions when I come back to work.

cats/js/src/main/scala/org/specs2/matcher/IOMatchers.scala
bpholt: I believe I did this correctly in b48c6cd, but please take a look

common/js/src/main/scala/org/specs2/reflect/Classes.scala
bpholt: Addressed in b48c6cd

bpholt:
I suspect the Scala 2.13 builds are going to fail… I couldn't figure out how to introduce cross-platform builds when we're overriding the Scala version for the cats artifacts. 

```scala
      Seq(libraryDependencies ++=
        (if (scalaVersion.value == "2.13.0-RC1")
           Seq("org.typelevel" % "cats-core_2.13.0-M5" % "1.6.0",
               "org.typelevel" % "cats-effect_2.13.0-M5" % "1.2.0")
         else
           Seq("org.typelevel" %%% "cats-core" % catsVersion,
               "org.typelevel" %%% "cats-effect" % catsEffectVersion)
        ))
```

I tried using 

```scala
      Seq(libraryDependencies ++=
        (if (scalaVersion.value == "2.13.0-RC1")
          crossVersion.value match {
            case cv: sbt.librarymanagement.Binary =>
               Seq("org.typelevel" % s"cats-core_${cv.prefix}2.13.0-M5${cv.suffix}" % "1.6.0",
                   "org.typelevel" % s"cats-effect_${cv.prefix}2.13.0-M5${cv.suffix}" % "1.2.0")
            case _ =>
               Seq("org.typelevel" % "cats-core_2.13.0-M5" % "1.6.0",
                   "org.typelevel" % "cats-effect_2.13.0-M5" % "1.2.0")
          }
```

Theoretically that should output the right artifact names, but sbt rejects it, complaining that I was trying to use conflicting cross-version suffixes:

```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/Users/bholt/Developer/github/specs2/"), "catsJS"):
[error]    org.scala-js:scalajs-library _2.13.0-RC1, _2.13.0-M5
```

I'm not sure how to resolve that short of waiting for cats to publish for 2.13-RC1 or -RC2?

etorreborre:
> I'm not sure how to resolve that short of waiting for cats to publish for 2.13-RC1 or -RC2?

I don't know how long it is going to take. I can see that only `cats-2.0.0-M1` is released for Scala 2.13. I suggest you send a message on gitter to ask the cats maintainers what we should expect there (because the 2019 roadmap document does not really say what's the plan for Scala 2.13)

bpholt:
It was pointed out to me in #cats-dev that there's no ScalaCheck 1.13 for Scala 2.13. AFAICT the main reason for cats/cats-effect going to 2.x is to update to ScalaCheck 1.14 (which is not binary-compatible with 1.13).

Since specs2 has already updated to ScalaCheck 1.14, projects that use cats-laws 1.x probably don't work now. We could just update to cats 2.x on Scala 2.13, but since this stuff is broken now anyway, I think it'd be better to upgrade to 2.x in general.

Both [cats-core](https://github.com/typelevel/cats/releases/tag/v2.0.0-M1) and [cats-effect](https://github.com/typelevel/cats-effect/releases/tag/v2.0.0-M1) have guaranteed the binary compatibility of their 2.x artifacts with the older 1.x artifacts, so projects not using ScalaCheck should be able to drop them in without issue.

bpholt:
Ok, I think this is ready for another review, @etorreborre. (The test failures are the same as the failures I'm seeing on master.) If this looks good, I can squash my commits before we merge to keep the history clean.

bpholt:
I think too much time has passed for this to realistically be merged, so I'll close it.

etorreborre:
Sorry I badly dropped the ball on that one. I am still working on specs2 5 for Scala3 where there will be no 3rd party dependencies (except maybe for Scalacheck) and maybe I will then revive your code as a separate `specs2-cats` library/project on Github.

PR #726, opened 2018-12-22
Update cats and cats-effects

xuwei-k:


etorreborre:
Thank you and have some good holidays 🎄!

PR #727, opened 2019-01-23
Gracefully shut down specs2 env on exception from JUnit's `Description` creation

hmemcpy:
This PR wraps the call to `JUnitRunner.getDescription` with a try/catch, calling `env.shutdown` before rethrowing the exception. This prevent external runners (bazel, in particular) from hanging indefinitely on a crash.

Details:
The following spec causes an exception to be thrown during `getDescription`:
```
class MySpec extends mutable.Specification {
    val boom: String = { throw new RuntimeException("boom!");
}
```
This happens, unfortunately, since mutable specs must be "executed" in order to create descriptions out of them.
Some runners (bazel's, in particular), may call `getDescription` before actual test run (before any `RunNotifier` is created), to perform some filtering.
It seems that some architectural change in v4.x of specs2 caused exceptions thrown from `getDescription` to keep the external runner "waiting", until it ultimately times out. This did not happen in v3.x of specs - exceptions from `getDescription` would properly terminate the runner.

Calling `env.shutdown` before rethrowing the exception from `getDescription` seem to solve the problem of the hanging test runner.

### DIFF
--- a/junit/shared/src/main/scala/org/specs2/runner/JUnitRunner.scala
+++ b/junit/shared/src/main/scala/org/specs2/runner/JUnitRunner.scala
@@ -12,6 +12,7 @@ import control._
 import org.specs2.fp.syntax._
 import org.specs2.concurrent.ExecutionEnv
 import ExecuteActions._
+import scala.util.control.NonFatal
 
 /**
  * Runner for specs2 specifications
@@ -35,8 +36,9 @@ class JUnitRunner(klass: Class[_]) extends org.junit.runner.Runner with Filterab
   lazy val env: Env =
     Env(arguments = arguments, lineLogger = LineLogger.consoleLogger)
 
-  lazy val getDescription =
-    JUnitDescriptions.createDescription(specStructure)(env.specs2ExecutionEnv)
+  lazy val getDescription: org.junit.runner.Description =
+    try JUnitDescriptions.createDescription(specStructure)(env.specs2ExecutionEnv)
+    catch { case NonFatal(t) => env.shutdown; throw t; }
 
   /** specification structure for the environment */
   lazy val specStructure: SpecStructure =


etorreborre:
Hi Igal, thanks for your PR :-)

Since the `Env` is created in the same class it makes sense to close it in the same place. The only way I see to test it is to write:
```
lazy val getDescription: org.junit.runner.Description =
    getDescription(env)

def getDescription(e: Env): org.junit.runner.Description =
    try JUnitDescriptions.createDescription(specStructure)(e.specs2ExecutionEnv)
    catch { case NonFatal(t) => env.shutdown; throw t; }
```
Then call the second method, and check that the underlying executor services in `ExecutionEnv` on the JVM are shutdown, possibly by bubbling up a `isShutdown` method from `ExecutorService` to `Env`.

I would probably rewrite things differently knowing more now about resources management with `ResourceT` and `bracket`...

hmemcpy:
Done :) I had to create a copy of the env to prevent `ownEnv` from shutting down prematurely. Thanks for the help!
Please let me know if I should do anything else?

junit/jvm/src/test/scala/org/specs2/runner/JUnitRunnerSpec.scala
```diff
@@ -135,3 +150,9 @@ class JUnitErrorInBeforeAllSpecification extends Specification with BeforeAfterA
 
   def afterAll: Unit = ()
 }
+
+class JUnitWithErrorInInitialization extends mutable.Specification {
+  val boom = { throw new RuntimeException("Error.") }
+
+  "example" >> ok
+}
```
etorreborre: Could you just add a newline here please?

etorreborre:
That's great, I just left a small comment. And then I will merge and publish a new version

junit/jvm/src/test/scala/org/specs2/runner/JUnitRunnerSpec.scala
hmemcpy: Done :)

ittaiz:
@etorreborre thanks! any idea when you'll be able to cut a new release?

hmemcpy:
Awesome, thanks so much Eric for the swift release!

PR #730, opened 2019-02-28
[WIP] Scala Native support. first commit

lolgab:
This is a work in progress PR about supporting Scala Native. Specs2 is a dependency of many other libraries. So it is important for the Scala Native ecosystem. Collaboration is welcomed. There are linking errors in parts where there are threads involved.


etorreborre:
Thanks @lolgab, do you have a good grasp of what needs to be done for the linking errors? Scala native is something I have absolutely never touched so I have no clue.

lolgab:
Scala Native is doesn't yet support multithread. So some tests need to be JVM only. And maybe some modules can't be ported (related to particular JVM libraries) so should be filtered out from the porting.
I give another look tomorrow :)

lolgab:
Pull Request superseded by https://github.com/etorreborre/specs2/pull/758.
I'm closing it.

PR #733, opened 2019-03-26
Remove sbt-buildinfo from root project

dragos:
The root project is only an aggregate of other projects and has no sources of its own. However, the buildinfo plugin creates one, that seems to be useless. This can have an impact on compilation time, but also confuses Hydra, and since this seems an oversight, the simplest is to remove it.


etorreborre:
No problemo, thanks!

PR #742, opened 2019-05-21
Upgrade to Scala 2.13.0-RC2

rossabaker:


build.sbt
```diff
@@ -120,7 +120,7 @@ lazy val cats = crossProject(JSPlatform, JVMPlatform).in(file("cats")).
   settings(
     moduleSettings("cats") ++
       Seq(libraryDependencies ++=
-        (if (scalaVersion.value == "2.13.0-RC1")
+        (if (scalaVersion.value == "2.13.0-RC2")
```
rossabaker: This is not legitimate, but not worse than what was here already.  I recommended to @bpholt earlier today that specs2 should be targeting cats-2.x and cats-effect-2.x, because 1.x on Scala 2.11 and 2.12 use laws based on scalacheck-1.13, whereas specs2 has already upgraded to scalacheck-1.14.  cats-2.x and cats-effect-2.x would get everything back on the same foundation, at the expense of being on milestones for a short while.

But that's a problem for another PR. :laughing:

etorreborre: Thanks @rossabaker. I'm fine to move to cats-2.x as long as users are trusting the milestones. Are there lots of projects already using them?

bpholt: FYI @rossabaker, I took your advice in #724, which addresses the cats upgrade, among other things.

rossabaker: Most of the cats-2.x I've seen are libraries preparing for Scala 2.13.  The problem is that specs2 is using scalacheck-1.14, and cats-2.x is where the laws catch up.  The good news is that cats-core-2.x is binary compatible with cats-core-1.x  

For apps, I'm still using a version of specs2 from before the scalacheck upgrade so it works with cats-laws-1.x.  I'm looking forward to all the versions being aligned again.

PR #746, opened 2019-06-10
update to Scala 2.13.0

kailuowang:


symbiont-eric-torreborre:
Thanks, I started looking at publishing, I'm just waiting on cats now, do you know when it will be ready?

kailuowang:
Thanks @symbiont-eric-torreborre I am working on cats. I need a specs2 release on 2.13 to unblock discipline, which blocks Cats. Unfortunately there is a circular dependency cats -> discipline -> specs2
 -> cats

etorreborre:
Oh, I didn't realize that. How do you break the cycle then? You publish specs2 locally, then discipline and cats for everyone? Maybe I should extract `specs2-cats` somewhere to avoid this

kailuowang:
TBC, It's just circular build dependency,  Cats does not depend on specs 2 jars.

kailuowang:
As we've done in the past, we publish specs2 4.5.1 2.13.0 , but it depends on Cats 1.6.0_2.13.0_M5, later, once Cats 2.0.0-RC1_2.13 is released we can update specs 2.

kailuowang:
I can publish specs2 locally as well, but then it won't pass the build for a discipline PR.

etorreborre:
Ok, so today you suggest that I publish specs2 with cats on `2.13.0-RC3`. I can do that

kailuowang:
@etorreborre yeah, thanks I think it's either cats-2.13.0-M5_1.6.0 or Cats_2.13.0-RC3_2.0.0-M3, I don't think we have cats 1.6.0 on Scala2.13.0-RC3

etorreborre:
It is published now

kailuowang:
Thanks very much!

Issue #750, opened 2019-06-28
[Feature Request] Ability to specify ScalaCheck parameters such as minSuccessfulTests  for each property


jkleckner:
Some properties of `ScalaCheck` are heavyweight and some less so.

It would be very nice to have a way to set the `minSuccessfulTests` and `maxDiscardRatio` inline with a test or at the class level.

symbiont-eric-torreborre:
Do you mean this: https://etorreborre.github.io/specs2/guide/SPECS2-4.3.4/org.specs2.guide.UseScalaCheck.html#test-properties?

jkleckner:
Uh, yes thanks.  The disconnect between `minTestsOk` as an argument and `minSuccessfulTests` as a parameter in `ScalaCheck` threw me off in my searching...

symbiont-eric-torreborre:
Yes that's my fault for preferring shorter names :-)

jkleckner:
@symbiont-eric-torreborre Not a problem.  Thank you very much for specs2.  I created a trivial MR #752 to make note of this.

PR #752, opened 2019-07-02
minTestsOk is minSuccessfulTests in ScalaCheck

jkleckner:
Added comments to guide and code noting that specs2 uses minTestsOk to
refer to the ScalaCheck minSuccessfulTests parameter.

See discussion in #750


etorreborre:
Thanks Jim

PR #754, opened 2019-07-10
Remove quote from test of XML escape

ashawley:
The plan in scala-xml 2.0 is to stop escaping quotation marks in XML element text (attribute text will still escape quotes).

XML element text is no longer being escaped with `scala.xml.Utility.escape`.  Scala XML could probably have something added to provide parity with the two escape situations, but I'm not sure there's much value.


etorreborre:
Thanks

PR #758, opened 2019-07-29
Scala native support

lolgab:


etorreborre:
@lolgab the build is passing I take it that it's ready to be merged?

lolgab:
Yes I'm OK with it. From my side you can merge it. If you can release a new version it could be great ☺

etorreborre:
Let's go, I'll do a release soon

etorreborre:
@lolgab I've had issues publishing the jars because Scala Native is only available for `2.11`. And when I publish for `2.12` sbt tries to re-publish the native jars for `2.11` which fails. Do you know the proper sbt way to avoid this?

lolgab:
It should not happen. I redefined `scalaVersion` and `crossScalaVersions` for native targets:
```scala
lazy val commonNativeSettings = Seq(
  scalaVersion := scala211,
  crossScalaVersions := Seq(scala211),
  nativeLinkStubs := true
)
```
Which command you use for publishing?

etorreborre:
`+publishSigned`

etorreborre:
I think we need to find a way to exclude those 2.11 modules for the publishing task, but I dread having to dig into sbt to find how :-)

lolgab:
Can you send me the sbt trace you get? even in PM.. I tried to publishLocal a toy project and it should exclude 2.12 versions for native target

symbiont-eric-torreborre:
If you do a `publishLocal` with `scalaVersion := "2.12.8"` you will see that the native modules are being published with a `2.11` version but this is not much trouble. However, when I use `+publishSigned` if some `2.11` modules have already been pushed to Sonatype because of the publication of specs2 for `2.11` before `2.12` or `2.13` I get errors because the same modules will be uploaded again and Sonatype won't allow it

lolgab:
It could be a bug of sbt-crossproject.
Look on how shapeless publishes its artifacts. They have separate tasks for JVM/JS and Native:
https://github.com/milessabin/shapeless/blob/master/scripts/travis-publish.sh
The root project aggregates only JVM and JS project.
https://github.com/milessabin/shapeless/blob/0cc1d28df5e09962ab4aa59402fc77d9e5560a66/build.sbt#L120

When I can I make a PR to remove artifact projects from root project and create another aggregate project for native target (if I can).
So you can do this:
```
+publishSigned
+native/publishSigned
```
or something like this.

symbiont-eric-torreborre:
That would be nice! As long as I don't have to go through each module and publish individually, having 2 steps is not that bad. Of course having scala native for 2.12 and 2.13 would be even better :-)

lolgab:
Scala Native for 2.12 and 2.13 is the goal for the next Milestone of 0.4.0
When it is out you can expect a new PR for another build clean-up :-)

PR #762, opened 2019-08-22
Fix website examples link

hagai-lvi:


etorreborre:
That's a good one, thanks!

Issue #764, opened 2019-09-06
Tests with multi-line names are skipped when using filter


mantasak:
I am facing an issue that test cases with multi-line names (e.g. using scala triple quote syntax; but not limited to) are being skipped. This only happens when using a test filter.

Given a test case:

```
import org.specs2.mutable.SpecificationWithJUnit

class QuickStartSpec extends SpecificationWithJUnit {

  "first\tsecond" >> {
    println("I am in tabbed test")
    1 must_== 1
  }

  "first\nsecond" >> {
    println("I am in newlined test")
    1 must_== 1
  }

}
```

The test with tab in its' name runs just fine with:

```
  sbt "testOnly QuickStartSpec -- -ex \Afirst\tsecond\Z"
```

However, I am unable to get the test with new line symbol to run in any way (it is always skipped with pass for the empty test suite).

I tried:

``` 
  sbt "testOnly QuickStartSpec -- -ex \Afirst\nsecond\Z"
```
```
  sbt "testOnly QuickStartSpec -- -ex \Afirst\A\Zsecond\Z"
```
``` 
 sbt "testOnly QuickStartSpec -- -ex \Qfirst\nsecond\E"
```
```
  sbt "testOnly QuickStartSpec -- -ex \Qfirst
  second\E"
```

As mentioned before, exact same behavior happens for tests that use multiline names with triple quote strings.

Can you please confirm if it is an issue with the regex filter, or is it my error?

Would be great to have such a case documented.

symbiont-eric-torreborre:
It is likely to be an issue with the regex filter yes. I'll have a look at that.

mantasak:
@symbiont-eric-torreborre I can help with verifying and/or fixing the issue if you want. 

In such case, I would appreciate if you could point me to the right direction where I can find the regex matching logic.

symbiont-eric-torreborre:
I would love to! 

The selection code starts [here](https://github.com/etorreborre/specs2/blob/master/core/shared/src/main/scala/org/specs2/specification/process/DefaultSelector.scala#L41).

[...]

PR #765, opened 2019-09-09
Allow matching special line characters with regex #764

mantasak:
This fixes the issue referred to in #764.


etorreborre:
Thanks a lot for this! I have a concern though. Some tests are now failing with
```
[error] x typecheck("val a: String = 1") must failWith("type mismatch")
[error]  type mismatch;
[error]   found   : Int(1)
[error]   required: String
[error]   doesn't match
[error]  type mismatch (TypecheckMatcherSpec.scala:14)
```
The reason is that `failWith(r)` expects a regular expression for `r` which will not care about newlines. So I would propose that you change the `failWith` matcher to suppress newlines there instead of doing it in `matchesSafely`.

mantasak:
I agree with you, will look into it. Couldn't get the entire repo to build locally successfully, so unfortunately did not notice the failing tests.

mantasak:
@symbiont-eric-torreborre I made the changes.

Please let me know if I understood your suggestion correctly.

etorreborre:
Yes all good thanks!

PR #766, opened 2019-09-17
Update scalacheck to 1.14.1-RC2. remove lolgab scalacheck

xuwei-k:


symbiont-eric-torreborre:
Nice one, thanks. Good to see that ScalaCheck is getting actively maintained again :-).

PR #767, opened 2019-09-18
Update dependencies

xuwei-k:


etorreborre:
Thanks for all these updates, maybe it's high time for me to add `specs2` to scala-steward. I just did it, let's see how this works out

PR #772, opened 2019-09-22
Allow to determine the duration before retry according the retry number

cchantep:
Usage:

```scala
eventually(retries = 3, _ * 100.milliseconds) {
  ???
}

foo must beTrue.eventually(3, _ * 75.milliseconds + 1.second)
```

### DIFF
--- a/common/shared/src/main/scala/org/specs2/execute/EventuallyResults.scala
+++ b/common/shared/src/main/scala/org/specs2/execute/EventuallyResults.scala
@@ -11,29 +11,45 @@ import scala.concurrent.duration._
 trait EventuallyResults {
 
   /**
+   * @param sleep the function applied on the retry number (first is 1)
    * @return a matcher that will retry the nested matcher a given number of times
    */
-  def eventually[T : AsResult](retries: Int, sleep: Duration)(result: =>T): T = {
-    def retry(retries: Int, sleep: Duration, r: =>T): T = {
-      if (retries <= 1) {
-        r
+  def eventually[T : AsResult](retries: Int, sleep: Int => Duration)(result: =>T): T = {
+    val max = retries - 1
+
+    @annotation.tailrec def retry(retried: Int): T = {
+      if (retried == max) {
+        result
       } else {
-        lazy val t = r
-        val result = ResultExecution.execute(t)(AsResult(_))
-        if (result.isSuccess)
+        lazy val t = result
+        val check = ResultExecution.execute(t)(AsResult(_))
+
+        if (check.isSuccess) {
           t
-        else {
-          Thread.sleep(sleep.toMillis)
-          retry(retries - 1, sleep, r)
+        } else {
+          val pause = sleep(retried).toMillis
+          Thread.sleep(pause)
+          retry(retried + 1)
         }
       }
     }
-    retry(retries, sleep, result)
+
+    if (retries <= 1) {
+      result
+    } else {
+      retry(0)
+    }
   }
+
+  /**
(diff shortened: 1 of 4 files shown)


common/shared/src/main/scala/org/specs2/execute/EventuallyResults.scala
```diff
@@ -11,29 +11,43 @@ import scala.concurrent.duration._
...
+  /**
+   * @return a matcher that will retry the nested matcher a given number of times
+   */
+  def eventually[T : AsResult](retries: Int, sleep: Duration)(result: =>T): T =
+    eventually[T](retries, (_: Int) => sleep)(result)
+
   /** @return a result that is retried at least 40 times until it's ok */
   def eventually[T : AsResult](result: =>T): T =
-    eventually(40, 100.millis)(result)
+    eventually(40, (_: Int) => 100.millis)(result)
```
etorreborre: You should be able to write `eventually(40, 100.millis)(result)` here, right?

common/shared/src/main/scala/org/specs2/execute/EventuallyResults.scala
```diff
@@ -11,29 +11,43 @@ import scala.concurrent.duration._
...
   /**
+   * @param sleep the function applied on the retry number (first is 1)
    * @return a matcher that will retry the nested matcher a given number of times
    */
-  def eventually[T : AsResult](retries: Int, sleep: Duration)(result: =>T): T = {
-    def retry(retries: Int, sleep: Duration, r: =>T): T = {
-      if (retries <= 1) {
-        r
+  def eventually[T : AsResult](retries: Int, sleep: Int => Duration)(result: =>T): T = {
+    @annotation.tailrec def retry(retried: Int): T = {
```
etorreborre: Good naming change 👍

matcher/shared/src/main/scala/org/specs2/matcher/EventuallyMatchers.scala
```diff
@@ -10,26 +10,35 @@ import org.specs2.execute.{ResultExecution, EventuallyResults}
...
   }
+  
+  /**
+   * @return a matcher that will retry the nested matcher a given number of times
+   */
+  def eventually[T](nested: =>Matcher[T], retries: Int, sleep: Duration): Matcher[T] = eventually[T](nested, retries, (_: Int) => sleep)
 
   /** @return a matcher that will retry the nested matcher 40 times  */
-  def eventually[T](nested: =>Matcher[T]): Matcher[T] = eventually(nested, 40, 100.millis)
+  def eventually[T](nested: =>Matcher[T]): Matcher[T] = eventually(nested, 40, (_: Int) => 100.millis)
```
etorreborre: Here you should also be able to call `eventually(nested, 40, 100.millis)` directly

common/shared/src/main/scala/org/specs2/execute/EventuallyResults.scala
cchantep: Yes, just prefer to directly call the more primitive variant

etorreborre: Ok, that's fine. On the other hand most of the tests are not passing. Could you please investigate why?

cchantep: Fixed

etorreborre:
All tests passing indeed, thanks!

cchantep:
Thx @etorreborre for the quick review

Issue #822, opened 2020-04-02
Fatal errors make test run forever


ivan-klass:
While using `org.specs2.mutable.SpecificationWithJUnit` (didn't checking other) fatal errors in tests (like missing jar dependency / incorrect jar version in classpath) make specification run forever, quick debug digging results in a loop within

[org.specs2.execute.ResultExecution#execute](https://github.com/etorreborre/specs2/blob/1acaab5aeeb5a5b0fc307bdcd9e846e690387cf4/common/shared/src/main/scala/org/specs2/execute/ResultExecution.scala#L21) -> 
[org.specs2.execute.ResultExecution#handleExceptionsPurely](https://github.com/etorreborre/specs2/blob/1acaab5aeeb5a5b0fc307bdcd9e846e690387cf4/common/shared/src/main/scala/org/specs2/execute/ResultExecution.scala#L26)

A very short example that makes test run forever is
```scala
import org.specs2.mutable.SpecificationWithJUnit

class Spec2ThrowableBug extends SpecificationWithJUnit {
  sequential

  def boom: String = throw new java.lang.NoSuchMethodError("Boom")

  "fatal stucks forever" >> {
    boom must beEqualTo("")
  }
}
```

etorreborre:
I think this is fixing the issue. Let's wait on CI to see if the build is green.

ivan-klass:
@etorreborre thank you very much, would you mind releasing a minor version with that fix?

etorreborre:
I just released `4.9.3` with the fix: https://repo1.maven.org/maven2/org/specs2/specs2-core_2.13/4.9.3/

PR #823, opened 2020-04-02
Fix fatal error issues (#822)

ivan-klass:
Not sure if it's a correct way to fix the issue, but I think that should help


etorreborre:
Yes I think it makes sense, thanks!

ivan-klass:
@etorreborre the failing CI point me to the spec that no fatal errors should be caught 
https://github.com/etorreborre/specs2/blob/2cfada84c67bc897919b31bf6e64b481ab11324b/core/jvm/src/test/scala/org/specs2/execute/ResultExecutionSpec.scala#L38-L39
The related commit message emphasize on catching **non-fatal**: https://github.com/etorreborre/specs2/commit/d568a9307c994a3f976b3905c08bc6b08be81ccf

I'm totally ok with that as soon as test runner will also throw the fatal error, not run forever. I still can't figure out why the execution stuck in a loop, could you please help?

etorreborre:
I'll investigate and come back to you soon.

ivan-klass:
Closed as out-dated, original #822 task is resolved

PR #826, opened 2020-04-24
fix #757

### LINKED ISSUE #757: Nested mutable examples not running with Scala 2.13
This spec here:

https://github.com/jroper/play-geojson/blob/master/src/test/scala/au/id/jazzy/play/geojson/ExampleSpec.scala

Has examples nested two deep (ie, `"foo" should { "do something" in { "bar" in { ... } } }`. Using specs 4.6.0, this runs fine with Scala 2.11 and 2.12, but with 2.13, it doesn't run at all, it's as if the specs aren't there. Not sure if this is a Scala issue, specs2 issue, or if it's by design with Scala 2.13 and I need to update my specs.

xuwei-k:
fix #757

maybe https://github.com/scala/scala/pull/7270 ? 🤔 

```scala
package example

import scala.reflect.NameTransformer

object A{
  Main.a = 1
}

object B{
  Main.b = 1
}

object Main {
  var a: Int = 0
  var b: Int = 0

  def initializeByConstructor(): Int = {
    this.a = 0
    val List(constructor) = Class.forName("example.A$").getDeclaredConstructors.toList
    constructor.setAccessible(true)
    constructor.newInstance()
    val result = Main.a
    this.a = 0
    result
  }

  def initializeByMODULE_field(): Int = {
    this.b = 0
    Class.forName("example.B$").getDeclaredField(NameTransformer.MODULE_INSTANCE_NAME).get(null)
    val result = Main.b
    this.b = 0
    result
  }

  def main(args: Array[String]): Unit = {
    println((A, B))
    println(scala.util.Properties.versionString)
    println(initializeByConstructor())
    println(initializeByMODULE_field())
  }
}
```

```
[info] version 2.12.11
[info] 1
[info] 0
```

```
[info] version 2.13.2
[info] 0
[info] 0
```

### DIFF
--- a/common/jvm/src/main/scala/org/specs2/reflect/Classes.scala
+++ b/common/jvm/src/main/scala/org/specs2/reflect/Classes.scala
@@ -1,7 +1,7 @@
 package org.specs2
 package reflect
 
-import scala.reflect.ClassTag
+import scala.reflect.{ClassTag, NameTransformer}
 import ClassName._
 import control._
 import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


etorreborre:
That looks good, thanks!

etorreborre:
@xuwei-k do you need a release with the fix?

xuwei-k:
Yes, please 🙏

etorreborre:
Done: https://repo1.maven.org/maven2/org/specs2/specs2-core_2.13/4.9.4/

PR #829, opened 2020-05-29
bump Scala versions (and be consistent about sbt version)

SethTisue:
this came up in the community build (failing tests)


SethTisue:
yay, CI is green now

PR #836, opened 2020-06-26
Try matcher fails with instance of Exception trait

cchantep:
When `Try` is handling an `Throwable` instanciated [from an anonymous subtype](https://github.com/etorreborre/specs2/pull/836/files#diff-aa179e2655318cda8e068701336d0888R58), matcher `beFailed.withThrowable[..](pattern)` is failing (due to the use of `haveClass` instead of `beAnInstanceOf`, as for the matcher `beFailed.withThrowable[..]` without message pattern).

See [test failing before](https://github.com/etorreborre/specs2/pull/836/files#diff-aa179e2655318cda8e068701336d0888R44): [Test result](https://travis-ci.org/github/etorreborre/specs2/jobs/702511734#L771)


cchantep:
Thx @etorreborre

PR #849, opened 2020-07-29
Fix catching unsupported step methods from IntelliJ

leedsalex:
IntelliJ has removed the Java prefix from their `Specs2Notifier` class.
https://github.com/JetBrains/intellij-scala/commit/d75e0c0e834dff9585e1cb0ced6d04e5c27248d3#diff-93cd505918cbca67189c8abea10d705aR19


etorreborre:
Thanks I didn't catch that!

PR #912, opened 2021-03-25
Depend on hamcrest instead of hamcrest-core which is deprecated

liucijus:
`org.hamcrest:hamcrest-core:2.2` has been deprecated and is an empty jar. For backwards compatibility `hamcrest-core` transitively brings `org.hamcrest:hamcrest:2.2`, which should be used instead.


etorreborre:
Thanks!

PR #946, opened 2021-05-29
docs: clarify behavior of foreach and forall matchers

MichalPawlicki:
Improve the documentation of the `foreach` and `forall` matchers. Followup to: https://stackoverflow.com/a/67750964/15227996


etorreborre (approved):
That's very cool, thank you!

PR #953, opened 2021-06-15
Upgrade kind-projector, replace deprecated `?` syntax

smarter:
The `?` syntax in types was deprecated and removed from kind-projector
in favor of `*`. `?` is reused for wildcards in Scala 3 as well as Scala
2 under `-Xsource:3`, and this meaning might also be supported by
default in a future version of Scala 2. See
http://dotty.epfl.ch/docs/reference/changed-features/wildcards.html for
more information.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -445,7 +445,7 @@ lazy val compilationSettings = Seq(
           "-Xlint:-byname-implicit")
     }
   },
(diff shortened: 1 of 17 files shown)


etorreborre:
Thanks @smarter! I have already made that move in my `specs2-five` branch (which should be the official specs2-5.0 soon I hope) but I did not realize this could be done with Scala 2

PR #979, opened 2021-08-30
Upgrade scala-xml to 2.0.1 for next 4.12.x release

mkurz:
@etorreborre What was the reason you reverted to scala-xml 1.3.0 here: https://github.com/etorreborre/specs2/commit/eee7bea1f81333d9c3561614fcd4e85b53a8383e?

### DIFF
--- a/project/depends.scala
+++ b/project/depends.scala
@@ -44,7 +44,7 @@ object depends {
       scalaParser.value
   }
 
-  def scalaXml = "org.scala-lang.modules" %% "scala-xml" % "1.3.0"
+  def scalaXml = "org.scala-lang.modules" %% "scala-xml" % "2.0.1"
 
   lazy val mockito  = "org.mockito"  % "mockito-core"  % "3.11.2"
   lazy val junit    = "junit"        % "junit"         % "4.13.2"


mkurz:
Also be aware I opened the PR against the `js-execution-context` branch, because it seems that is the branch the last `4.12.3` release was cut from (and also that `4.12.4-js-ec` release)

mkurz:
@etorreborre I just saw that you upgraded to scala-xml 2.0.1 in https://github.com/etorreborre/specs2/commit/631f40575cdbbb3497116299903acd2d128a564d and released specs 4.12.4
Actually I am not sure anymore if we really should upgrade to scala-xml 2.0.1 in the 4.12.x branch (that's why a closed this PR yesterday night again). Because I realized that you will not publish specs 4.x for Scala 3, but only 5.x (currently RC phase).
So there is actually not really a point to upgrade scala-xml to latest in the 4.x branch anymore. The problem now however is that people that use sbt 1.5 will run into version conflicts with 4.12.4 now because suddenly scala-xml was bumped to a new major version. So I think it might be a good idea to downgrade to 1.3.0 again and release 4.12.5? WDYT?

etorreborre:
Yes, after I did the work to separate the `scala-xml` dependency from `scala-compiler` in specs2 I read your comments on `scala/scala#9743` seeing that it was a can of worms.

> Because I realized that you will not publish specs 4.x for Scala 3, but only 5.x (currently RC phase)

That's correct.

> The problem now however is that people that use sbt 1.5 will run into version conflicts with 4.12.4 now because suddenly scala-xml was bumped to a new major version

What I did not realize with `4.12.4` was that moving the `scala-xml` dependency to `specs2-xml` was a bit like hiding the issue under a rug since that still means that if you want to depend on both `specs2-matcher-extra` (which uses `scala-compiler`) and `specs2-xml` (which depends on `scala-xml`) you will get a conflict. So for that reason alone I am going to push out a `4.12.5` reverting the `scala-xml` upgrade and only use `scala-xml-2.0.1` for the `5.x` series.

mkurz:
> So for that reason alone I am going to push out a `4.12.5` reverting the `scala-xml` upgrade and only use `scala-xml-2.0.1` for the `5.x` series.

Yes, I think that is the way to go.

etorreborre:
Done.

mkurz:
@etorreborre Thanks!

PR #981, opened 2021-09-01
Regression when using .pendingUntilFixed + BeforeAll

melezov:
A reproducible example on something that's exploding our tests since we upgraded from 4.12.3
Tests that use .pendingUntilFixed and BeforeAll regress with Specs2 exploding on what seems to be valid code

    [error] cannot create an instance for class ExampleSpec
    [error]   caused by org.specs2.matcher.MatchFailureException: false != true
    [error]   caused by java.lang.Exception: false != true
    [error]   java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)

Specs2 versions from 4.12.4 to 4.12.6 are affected

### DIFF
--- a/core/jvm/src/test/scala/org/specs2/execute/PendingUntilFixedBeforeAllSpec.scala
+++ b/core/jvm/src/test/scala/org/specs2/execute/PendingUntilFixedBeforeAllSpec.scala
@@ -13,6 +13,6 @@ class PendingUntilFixedBeforeAllSpec extends Specification with BeforeAll {
 
   "Pending until fixed can be used with BeforeAll" >> {
     setupFinished ==== true
-    (1 ==== 1).pendingUntilFixed("Will explode")
+    (1 ==== 2).pendingUntilFixed("Will explode")
   }
 }
--- a/core/jvm/src/test/scala/org/specs2/execute/PendingUntilFixedBeforeAllSpec.scala
+++ b/core/jvm/src/test/scala/org/specs2/execute/PendingUntilFixedBeforeAllSpec.scala
@@ -1,0 +1,18 @@
+package org.specs2
+package execute
+
+import org.specs2.mutable.Specification
+import org.specs2.specification.BeforeAll
+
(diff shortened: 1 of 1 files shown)


etorreborre:
Thanks for the reproducible use case. The issue I am seeing here is with an implicit conversion.

If we write
```scala
"Pending until fixed can be used with BeforeAll" >> {
  setupFinished ==== true
  (1 ==== 2).pendingUntilFixed("Will explode")
}
```

Then it is more or less converted to:
```scala
"Pending until fixed can be used with BeforeAll".>>({
  setupFinished ==== true
  toExecution((1 ==== 2).pendingUntilFixed("Will explode"))
})
```
Which means that the implicit which transforms the block between `>>{...}` only applies to the last statement in order to create an `Execution` which is normally executed later on when the specs executes. But with the form above `setupFinished ==== true` is a statement which gets executed right away _when the specification object is being instantiated_. And `setupFinished` at this stage is `false` because the `beforeAll` method did not have a chance to execute. Then, since `====` throws an exception, this means that the specification itself cannot even be constructed.

I have a fix on the way and a `4.12.7` version.

etorreborre:
I'm closing the PR since I added a example to an existing spec for this behaviour. When `4.12.7` is published please try it out and tell me if it works ok for you.

melezov:
I'm running tests now, early warning that we seem to be seeing new types of errors:
```class org.specs2.specification.core.Fragment cannot be cast to class org.specs2.specification.core.Execution```

**Comment by melezov** [comment]
Here is the minimal failing spec and the accompanying error:

```class ErrorPathSpec extends org.specs2.mutable.Specification {
  "Minimal Test" >> {
    "Nested" >> {
      1 ==== 1
    }
  }
}
```

Test log:
```[info] ErrorPathSpec
[error] ! Minimal Test
[error]  java.lang.ClassCastException: class org.specs2.specification.core.Fragment cannot be cast to class org.specs2.specification.core.Execution (org.specs2.specification.core.Fragment and org.specs2.specification.core.Execution are in unnamed module of loader 'app') (file:1)
[```

PR #988, opened 2021-09-09
Use cross directories for cross projects

armanbilge:
Just a minor gripe, but projects with platform-specific sources aren't `CrossType.Pure` and use a `shared/`, `jvm/`, `js/` directory structure. (The `.jvm` and `.js` hidden directories for `CrossType.Pure` are more of an internal implementation detail.)


etorreborre:
Thanks a lot. I must say I'm not very fluent with cross-building, especially because I am not a ScalaJS user. I'll wait until the build is green to merge.

Also I realized that the formatting and mima checks are running way too many times. I need to fix that.

armanbilge:
Yeah, cross-building is still fairly messy. Recently we're motivated to use ScalaJS for serverless (e.g., see https://github.com/typelevel/feral) so we're pushing more on it.

I noticed that your tests are JVM-only, unless I'm missing something I think it would be good to get at least the cross-platform ones running on JS as well?

etorreborre:
> I noticed that your tests are JVM-only

Oh really? Where do you see that? In that case yes I want to enable them for both platforms

armanbilge:
https://github.com/etorreborre/specs2/blob/main/build.sbt#L185

They're just a `project`, not a cross project. I'm working on a PR for this. Thanks for merging!

etorreborre:
Ah, I see, those are tests that are depending on several other modules. So we indeed have ScalaJS tests for, say `specs2-core`, but not for those additional tests. I think I removed the cross project because some of those tests had to be classified as jvm vs js. Thanks for the PR!

armanbilge:
Ah yes you are absolutely right. Although, I had to add `ScalaJSJUnitPlugin` to `specs2-core` to get those tests running on my local, they're definitely running in CI? 🤔

But yes, I'm trying to classify those additional tests as JVM-only or shared so we can get a little more coverage for JS if possible :)

etorreborre:
I'm using `sbt` to run tests, and for ScalaJS tests I'd run something like:
```
specs2(main)> coreJS/testOnly *
```
On CI, I'm using `testOnly -- xonly exclude ci,website timefactor 3` which should run all the JVM and JS tests. If I look at the `AnsiColorSpec` for example I can see that [it runs twice](https://github.com/etorreborre/specs2/runs/3556363485?check_suite_focus=true).

armanbilge:
Yes, I must either be confused or mucked up something on my local, sorry about that.

PR #989, opened 2021-09-09
Get tests project running on JS

armanbilge:
@etorreborre With some reshuffling I got about half of the `tests` project running on JS. Not everything is passing at the moment, I was hoping maybe you could help with this part?

```
[error] Error: Total 616, Failed 44, Errors 5, Passed 567
[error] Failed tests:
[error]         org.specs2.scalacheck.ScalaCheckMatchersResultsSpec
[error]         org.specs2.matcher.ExceptionMatchersSpec
[error]         org.specs2.matcher.LogicalMatcherSpec
[error]         org.specs2.matcher.TryMatchersSpec
[error]         org.specs2.matcher.NumericMatchersSpec
[error]         org.specs2.matcher.BeEqualToMatcherSpec
[error]         org.specs2.matcher.ResultMatchersSpec
[error] Error during tests:
[error]         org.specs2.matcher.StringMatchersSpec
[error]         org.specs2.matcher.TraversableMatchersSpec
[error]         org.specs2.scalacheck.ScalaCheckMutableSpec
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
```


armanbilge:
Cool, I enabled ES2018 and got it down to:
```
[error] Error: Total 616, Failed 4, Errors 3, Passed 609
[error] Failed tests:
[error]         org.specs2.scalacheck.ScalaCheckMatchersResultsSpec
[error]         org.specs2.matcher.NumericMatchersSpec
[error] Error during tests:
[error]         org.specs2.matcher.TraversableMatchersSpec
[error]         org.specs2.scalacheck.ScalaCheckMutableSpec
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
```

etorreborre:
I fixed `ScalaCheckMatchersResultsSpec`, 2 examples from `NumericMatchersSpec` but I have some doubts over the 3rd failing example (`0.0 must not(beCloseTo(0.1 `within` 1.significantFigure))`, I wonder if there's a catch with numbers and JavaScript compared to the JVM).

The last 2 specs2 seems to be failing because of we are awaiting a future. I'll see if I can change the code to use futures only.

armanbilge:
Nice, thanks!! Yes, JS has _really_ weird semantics around numbers because at runtime everything is essentially a `Double`. See https://www.scala-js.org/doc/semantics.html. Is there a way we can work around this?

> I'll see if I can change the code to use futures only.

Amazing!! That would be ideal. Eventually, I hope we can get more of the tests running on JS. For example `AnyMatchersSpec` seems to be using JVM-only classes in the examples (e.g. `BufferedInputStream`) which was problematic :)

I've found tests are not just important for correctness but also for demonstrating that certain code can be used from JS without running into weird linking issues.

etorreborre:
I have reworked a few things like the `ActionMatchers` (matchers for the `Action` data type which is internal to specs2) and how the `Env` is being shutdown so that 2 more specs are fixed. I will need to sleep on it and check what to do with the `Numeric` failure

etorreborre:
All the tests are passing now. Note that I didn't really fix the `NumericMatchers` failures because there is a real difference between the value of the double `0.0` on the JVM and on JS 🤷.

armanbilge:
Thank you so much for your help!! Yep, JS numbers be that way 😕 If you're happy with this, I think it's good to go.

etorreborre:
Great, thanks for the final validation

etorreborre:
And thanks for starting this PR in the first place ❤️ !

armanbilge:
At some point I'd like to circle back and try to get more tests running on JS. But next stop, specs2-cats 😄

PR #992, opened 2021-09-10
Make ExecutionEnv consistent across JVM/JS

armanbilge:
This makes the `ExecutionEnv` API consistent across JVM and JS. Came up while working on https://github.com/etorreborre/specs2-cats/pull/4.


etorreborre:
💯

PR #997, opened 2021-09-21
sbt-scoverage 1.9.0 (was 1.8.2)

SethTisue:
this paves the way for an eventual upgrade to sbt 1.6 (1.6.0-M1 doesn't work with 1.8.2, because Scala 2.12.15)

it would be helpful to me in the Scala community build if this were merged


etorreborre:
@SethTisue done

PR #1003, opened 2021-10-05
fix scala 3 syntax at index page

unkarjedy:
fix scala 3 syntax at index page: remove closing brace in indentation based syntax


etorreborre:
Thanks a lot for this fix!

PR #1061, opened 2022-02-12
4.x branch: version bumps (sbt, Scala, sbt-github-actions); and fix CI to test what it should

SethTisue:
I initially tried also adding JDK 17 to the CI matrix, but of course that fails because of the usual parboiled2 issue.


SethTisue:
(Note that I'm not sure if publishing is right, but let's at least get CI right first before worrying about publishing.)

SethTisue:
also for all I know the Scala Native build is broken, but if it's being tested in CI anyway, then 🤷

SethTisue:
gah... one of the jobs has been running for 4 hours, but I don't think there's a way to see where it's stuck. I guess it will time out and then we'll see? or I could cancel it, but I guess waiting is fine.

SethTisue:
well, I only embarked on this on the first place because I thought it would be easy 🤷

PR #1067, opened 2022-02-19
Remove bad trailing comma

adampauls:
There is a [PR](https://github.com/lampepfl/dotty/pull/14517) up to remove bad trailing commas from dotty. This bad trailing comma is breaking the community build.


etorreborre:
Thanks!

PR #1097, opened 2022-08-08
add missing expression for try matcher

kanitkameh:
I saw on https://etorreborre.github.io/specs2/guide/SPECS2-3.4/org.specs2.guide.Matchers.html the line:
`beSuccessfulTry.withValue(exp) check if an element is Success(_)` which I guess is a mistake since the method we call is called `withValue` and has an expression as an argument.


etorreborre:
That's correct, thanks for the PR!

PR #1103, opened 2022-09-26
Upgrade scala-xml to 2.1.0 for next 4.17 release

mkurz:
I know once you already upgraded to scala-xml for 4.x, but [later reverted that again](https://github.com/etorreborre/specs2/commit/eee7bea1f81333d9c3561614fcd4e85b53a8383e), we also had [a discussion](https://github.com/etorreborre/specs2/pull/979).

However, the Scala world is moving towards scala-xml 2.x, even for current stable releases
* Scala 2.12.17 upgraded to scala-xml 2.1.0: https://github.com/scala/scala/pull/10108
* The next sbt version will then come with Scala 2.12.17: https://github.com/sbt/sbt/pull/7021
* sbt-native packager upgraded in its 1.9.10 (patch) release: https://github.com/sbt/sbt-native-packager/pull/1514
* More libraries are upgrading: https://github.com/sbt/sbt-native-packager/issues/1513
* I am in the processing of upgrading libraries I am maintaining, e.g. https://github.com/playframework/twirl/pull/547

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.16]
+        scala: [2.13.8, 2.12.17]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.16]
+        scala: [2.13.8]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.17]
+        scala: [2.12.16]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.17]
+        scala: [2.12.16]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.15]
+        scala: [2.12.17]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.15]
+        scala: [2.12.17]
         java: [adopt@1.8]
     runs-on: ${{ matrix.os }}
     steps:
--- a/build.sbt
+++ b/build.sbt
@@ -33,9 +33,9 @@ lazy val specs2Settings = Seq(
   organization := "org.specs2",
   GlobalScope / scalazVersion := "7.2.32",
   specs2ShellPrompt,
-  scalaVersion := "2.13.8",
+  ThisBuild / scalaVersion := "2.13.8",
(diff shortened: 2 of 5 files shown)


etorreborre:
Since Scala 2.12.17 is making the move to scala-xml-2.0 too that makes sense. I will do a release later on today.

etorreborre:
I started the workflow. You need to regenerate the `ci.yml` file. I did it [on my branch](https://github.com/etorreborre/specs2/commit/82e7ebcb412c822425e6cac3a168c6dfb71968c5) if you want to cherry-pick the commit

mkurz:
OK I found out: If I upgrade to sbt 1.7.1, `githubWorkflowGenerate` wants to set the workflow versions to 2.12.16. So it sets the versions that sbt uses.

mkurz:
Alright, so I upgraded sbt as well and Scala now, I hope that works.

mkurz:
@etorreborre Can you please approve the workflow?

Also, I upgraded sbt/scala in the cross branch: #1104, for the sake of completeness

mkurz:
:cry: now we have the problem already:
```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error] 
[error] 	* org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over 1.0.6
[error] 	    +- org.specs2:specs2_2.12:4.16.1                      (depends on 2.1.0)
[error] 	    +- org.scala-lang:scala-compiler:2.12.16              (depends on 1.0.6)
```

so we need to tell `sbt-github-actions` somehow to use Scala 2.12.17...

etorreborre:
I think we should try to force `2.12.17`. I can have a look at that later on today if you want.

mkurz:
@etorreborre I think I found the solution. sbt-github-actions' [README says](https://github.com/djspiewak/sbt-github-actions):

> ... with sbt-github-actions, the `scala:` entry in the job `matrix:` is populated from the `ThisBuild / crossScalaVersions` key in your **build.sbt**.

Further [down below](https://github.com/djspiewak/sbt-github-actions):

> Any and all settings which affect the behavior of the generative plugin should be set in the `ThisBuild` scope (for example, `ThisBuild / crossScalaVersions :=` rather than just `crossScalaVersions := `). This is important because GitHub Actions workflows are global across the entire build, regardless of how individual projects are configured.

So that's what I did, see the last commit, I added `ThisBuild / ...`

What do you think?

etorreborre:
Let's see what the workflow says :-)

.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.15]
+        scala: [2.13.8]
```
mkurz, line 64: I think it does not really matter if using 2.12 or 2.13 here, since the publish is done cross anyway, see `run: ...` below: `sbt ++${{ matrix.scala }} +publish` If setting ++2.12.17 or ++2.13.8 shouldn't matter since publish is done with `+*.

build.sbt
```diff
@@ -33,9 +33,9 @@ lazy val specs2Settings = Seq(
   organization := "org.specs2",
   GlobalScope / scalazVersion := "7.2.32",
   specs2ShellPrompt,
-  scalaVersion := "2.13.8",
+  ThisBuild / scalaVersion := "2.13.8",
   SettingKey[Boolean]("ide-skip-project").withRank(KeyRanks.Invisible) := platformDepsCrossVersion.value == ScalaNativeCrossVersion.binary,
-  crossScalaVersions := Seq(scalaVersion.value, "2.12.15"))
+  ThisBuild / crossScalaVersions := Seq("2.13.8", "2.12.17"))
```
mkurz, line 38: So, if you use `scalaVersion.value` inside `crossScalaVersions` and you run the whole thing with `++2.12.7`, you will then end up with `Seq("2.12.17", "2.12.17")`. This just bit me when running `githubWorkflowCheck`, because it will check for `scala: [2.12.17, 2.12.17]` inside `ci.yml` which of course is nonsense and will always fail the check...

etorreborre:
I invited you as a collaborator to make things easier

mkurz:
@etorreborre thanks for inviting me! Finally this pull request turned green. Can we merge and release 4.16? Thanks!

etorreborre:
Thanks for your hard work!

etorreborre:
I'm on it :-)

PR #1112, opened 2022-10-27
Update AnyMatchers.scala

dekelatwix:
This fixes a problem where you can't use beEqualTo as a parameter matcher when mocking in mockito.

PR #1116, opened 2022-11-02
revert #1112

dekelatwix:
https://github.com/etorreborre/specs2/pull/1112/files This causes beEqualTo usages with vague types to fail compilation.

### DIFF
--- a/tests/shared/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
+++ b/tests/shared/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
@@ -34,6 +34,7 @@ class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers 
   ${"a" must not(be_==("b"))}
   ${"a" must be_!=("b")}
   ${"a" must not(be_!=("a"))}
+  ${object A {override def equals(that: Any) = that == "a"}; A must be_==("a")}
 
   Array equality uses deep array comparison, with or without typed equality
   ${Array(1, 2) must be_==(Array(1, 2))}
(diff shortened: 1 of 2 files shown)


dekelatwix:
@etorreborre Yeah, I wasn't very smart with that one. 
The reason for BeEqualTo to behave the old way is written here. https://github.com/etorreborre/specs2/blob/76e79acc3f33a32ed14f1621cb62d26c1726f1a4/notes/1.3.markdown#L11
Sorry for the hassle

etorreborre:
Hmm, that doesn't look good on me to approve a PR that actually break things 😬. I had a vague feeling about that in my head but didn't have the time to really dig in. Maybe one thing that I can do to avoid this regression in the future would be to use [a Typecheck matcher](https://etorreborre.github.io/specs2/guide/SPECS2-5.0.7/org.specs2.guide.Matchers.html#typecheck-matchers) to show what is not supposed to compile. Do you have a small example showing this case? Can you add a small example, for example in [this spec](https://github.com/etorreborre/specs2/blob/main/tests/shared/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala)? (if you just paste a small example on this issue I can also add the test myself, no worries).

dekelatwix (commented):
added test

tests/shared/src/test/scala/org/specs2/matcher/BeEqualToMatcherSpec.scala
```diff
@@ -6,6 +6,9 @@ import org.specs2.text.AnsiColors.*
 
 class BeEqualToMatcherSpec extends Spec with ResultMatchers with ShouldMatchers {
   def is = s2"""
+  object A {
```
etorreborre, line 9: The object needs to be defined outside of the `s2""" ... """` block because otherwise it is just a piece of text :-).

dekelatwix:
Did you like the test? Do you have an idea to reduce the code size of it? @etorreborre

etorreborre:
The code size is fine with me but could you please move the object definition towards the end of the file with the rest of the support code? I generally try to put the examples first and then all the support code in specifications.

dekelatwix:
How about this version? 🕺

etorreborre:
Great, thanks for the clean-up!

PR #1128, opened 2023-01-18
backport JsonMatchers as companion object

nadavwe:
Hi @etorreborre,
we have quite a lot of places using just a few matchers from JsonMatchers.
I thought having them in object form would be a nice addition and saw that it already exists in mainline.
can it be backported? (I hope it's on the right branch...)
Thanks!
(this PR is just for the discussion, feel free to close/edit/whatever...)


nadavwe:
thanks! :)

PR #1152, opened 2023-05-07
Update correct line numbers in LocationSpec

som-snytt:
The test breaks with https://github.com/lampepfl/dotty/pull/17428

I'm not sure what the protocol is for a project participating in the community build, where this behavior change was noticed.

Parking this here for reference.


etorreborre:
Thanks @som-snytt. I adjusted in the fix in https://github.com/etorreborre/specs2/commit/947e823df746315a07f556c29172b300117d4a40 so that the spec passes both specs2 current CI and the community build for the next scala version.

etorreborre:
One think I don't know though is how the community build selects a proper specs2 version for its build.

som-snytt:
Clever! Thanks, and of course I have no idea if/when the PR might be accepted on the Scala 3 side.

PR #1153, opened 2023-05-10
Specify this.run in run

som-snytt:
The new comment:

It is an error if an identifier x is available as an inherited member in an inner scope and the same name x is defined in an outer scope in the same source file.

`run` also has an `apply`, so `run()` is doubly potentially confusing.

Just slapping on a disambiguating `this` is harmless and also clarifying.


etorreborre:
Thanks!

PR #1161, opened 2023-06-03
feat: Add notThrow matcher

tmccombs:
This is a matcher that checks that no exception is thrown.

It has two advantages over `not(throwA[Throwable])`:

1. It includes the stacktrace of the original exception in the message if it fails.
2. It doesn't require specifying a type.


tmccombs:
Alternatively the `not` override on ExceptionClassMatcher could be modified to include the stack trace. However, I think that would be significantly more complicated, and detecting any exception is usually what you want anyway.

etorreborre:
Hi @tmccombs, if you don't mind you PR actually prompted me to go one step further and implement the full behaviour for `not(throwAn[Exception])`. I think it is more consistent with the rest of the library than introducing an ad-hoc `notThrow` matcher

tmccombs:
Awesome, thank you

PR #1163, opened 2023-06-08
Add stack trace to output for beSuccessfulTry on failure

tmccombs:
This isn't strictly backwards compatible, because the success matchers no longer extend OptionLike matchers. However, i'm not sure if that would break anything in practice. 

If we want to keep full backwards compatibility I can change it to use the same base classes as before.

### DIFF
--- a/matcher/shared/src/main/scala/org/specs2/matcher/TryMatchers.scala
+++ b/matcher/shared/src/main/scala/org/specs2/matcher/TryMatchers.scala
@@ -47,16 +47,16 @@ trait TryMatchers:
 object TryeMatchers extends TryMatchers
 
 case class TrySuccessMatcher[T]() extends Matcher[Try[T]]:
-  def apply[S <: Try[T]](value: Expectable[S]) = checkSuccess(value, ValueCheck.alwaysOk)
+  def apply[S <: Try[T]](value: Expectable[S]): Result = checkSuccess(value, ValueCheck.alwaysOk)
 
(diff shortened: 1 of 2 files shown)


matcher/shared/src/main/scala/org/specs2/matcher/TryMatchers.scala
```diff
@@ -45,11 +46,17 @@ trait TryMatchers:
 
 object TryeMatchers extends TryMatchers
 
-case class TrySuccessMatcher[T]() extends OptionLikeMatcher[Try[T], T]("a Success", (_: Try[T]).toOption):
+case class TrySuccessMatcher[T]() extends Matcher[Try[T]]:
+  def apply[S <: Try[T]](value: Expectable[S]) = checkSuccess(value, ValueCheck.alwaysOk)
```
etorreborre, line 50: If you don't mind, I'd like to take the opportunity, since we're touching that code, to add return types to the methods. I've been guilty and not doing that systematically when using Scala and I'd rather do it now.

etorreborre (commented):
That looks great, thanks! If you don't mind putting the return types on methods, then let's merge it.

matcher/shared/src/main/scala/org/specs2/matcher/TryMatchers.scala
etorreborre: No, this is the just the very old Scala newbie-me who was like "no return types needed, woohoo!". I've learned better since but never came back to fix everything.

etorreborre:
Thank you!

PR #1216, opened 2024-01-19
Filter examples based on TestSelector

Duhemm:
Previously, the Specs2 runner would ignore the selectors that are passed via a `TaskDef`.

With this patch, when a `TaskDef`'s selectors contains only `TestSelector`s, then only the examples whose name match the input selectors will be executed.

### DIFF
--- a/core/shared/src/main/scala/org/specs2/runner/SbtRunner.scala
+++ b/core/shared/src/main/scala/org/specs2/runner/SbtRunner.scala
@@ -16,6 +16,9 @@ import org.specs2.fp.*, syntax.*
 import org.specs2.concurrent.ExecutionEnv
 import org.specs2.data.NamedTag
 import scala.util.*
+
+import java.util.regex.Pattern
+
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, Future, ExecutionContext}
 
@@ -37,7 +40,14 @@ abstract class BaseSbtRunner(args: Array[String], remoteArgs: Array[String], loa
 
   /** create a new test task */
   def newTask(aTaskDef: TaskDef): Task =
-    SbtTask(aTaskDef, env, loader, this)
+    val fullEnv =
+      if (env.arguments.select._ex.isDefined || aTaskDef.selectors().exists(!_.isInstanceOf[TestSelector])) env
+      else
+        val names = aTaskDef.selectors().toList.collect { case ts: TestSelector => Pattern.quote(ts.testName()) }
+        val select = env.arguments.select.copy(_ex = Some(names.mkString("|")))
+        env.setArguments(env.arguments.copy(select = select))
+
+    SbtTask(aTaskDef, fullEnv, loader, this)
 
   def done =
     env.shutdown
--- a/core/jvm/src/test/scala/org/specs2/runner/SbtSelectorSpec.scala
+++ b/core/jvm/src/test/scala/org/specs2/runner/SbtSelectorSpec.scala
@@ -1,0 +1,84 @@
+package org.specs2
+package runner
+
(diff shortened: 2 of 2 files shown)


Duhemm:
In case this ends up being accepted, I've already backported this to 4.x: [backport-test-selector](https://github.com/Duhemm/specs2/tree/backport-test-selector)

etorreborre:
Hi Martin. First of all, thanks for this PR, it looks great with all the tests! However I don't get what is the use for a `TestSelector` with `sbt` on the command line. I have always selected classes with a wildcard like `testOnly *MySpec*` and then individual tests (or examples in BDD terms) with `testOnly *MySpec* -- ex "first example"`. How can I create a `TestSelector` from sbt's command line?

etorreborre:
Confession: I just discovered that sbt was proposing to complete the spec names for the `testOnly` command when pressing `TAB`. Better late than never :-).

Duhemm:
Hi Eric, thanks for taking a look.

On the command line, I don't know of a way to use the selectors. The selectors are intended to be used programmatically, for instance by an IDE that would trigger the execution of a single test, or an sbt plugin that would retry failed tests.

In the examples that you showed, the syntax to execute only a single test case works only for Specs2, whereas the solution based on test selectors is framework-agnostic.

My understanding is that this is the reason why, in sbt's test API, `TaskDef` has a list of selectors: so that it can be used to specify which test cases to execute within a test suite. I've noticed that ScalaTest, JUnit and MUnit already support filtering test cases based on selectors (I have pull requests open for the other major test frameworks).

Please let me know if this makes the intent behind this pull request clearer.

core/shared/src/main/scala/org/specs2/runner/SbtRunner.scala
```diff
@@ -37,7 +40,14 @@ abstract class BaseSbtRunner(args: Array[String], remoteArgs: Array[String], loa
 
   /** create a new test task */
   def newTask(aTaskDef: TaskDef): Task =
-    SbtTask(aTaskDef, env, loader, this)
+    val fullEnv =
+      if (env.arguments.select._ex.isDefined || aTaskDef.selectors().exists(!_.isInstanceOf[TestSelector])) env
```
etorreborre, line 44: One question about this condition: shouldn't it be `aTaskDef.selectors().all(!_.isInstanceOf[TestSelector])`, meaning "don't do anything if there's no test selector"?

etorreborre:
Thanks for your explanation @Duhemm. I left another question about the code because I'm not sure about one condition.

core/shared/src/main/scala/org/specs2/runner/SbtRunner.scala
etorreborre: @Duhemm ☝️ in case you didn't see the message.

Duhemm: Thanks for the reminder 😄  I don't think so. The idea was to revert to the current behavior unless we have only `TestSelector`s.

Otherwise, I'm not sure what would be the best behavior when we receive a `SuiteSelector` (ie. run all tests) and a `TestSelector` (ie. run a single test). There are also other types of selectors that are not used in Specs2.

etorreborre:
Excellent, thanks @Duhemm. Can you please open a PR for the backport as well?
Also, how fast do you need a release for this?

Duhemm:
Thanks, I just opened #1217.

There's no rush in releasing! Sometime before March would be great though 😄

etorreborre:
Perfect, that will definitely be done by then :-)

etorreborre:
This is published now as `4.20.5` for the maintenance branch and `5.5.1` for the main branch.

PR #1241, opened 2024-06-02
scala-native 0.5 for `specs2-cross` branch

xuwei-k:


etorreborre:
This is published as `4.20.7`

PR #1246, opened 2024-06-08
chore: remove unused paradise dependency

NTPape:
It was only used for Scala <= 2.10, so has been unused for a while now.


etorreborre:
Makes sense, thanks!

PR #1247, opened 2024-06-09
chore: clean up inter-project dependencies

NTPape:
Remove unused dependencies between the specs2 modules. This change is "breaking" in the sense that projects using the specs2 library may have to be more explicit with their specs2 dependency declarations; most notably `specs2-core`, `specs2-analysis` and `specs2-matcher-extra` may need to be added explicitly if e.g. only `specs2-html`, `specs2-gwt` or `specs2-mock` were depended on before.

This change originates from the observation that `html` currently pulls in a `scala-compiler` dependency, which came via  (`form`,) `matcherExtra` and `analysis`, which are all not actually needed for `html`. The same holds true for `form` and `gwt`.

For a full list of changes, refer to the diff.

(I don't know if this kind of "breaking change" is acceptable. If not, feel free to close this PR. If only some of these changes are wanted, I can change it accordingly. I will wait for feedback before making a similar change for the master branch.)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -88,7 +88,7 @@ lazy val analysis = crossProject(JSPlatform, JVMPlatform, NativePlatform).in(fil
     commonJvmSettings).
   jsSettings(commonJsSettings).
   nativeSettings(commonNativeSettings).
-  dependsOn(common % "test->test", core, matcher, scalacheck % Test)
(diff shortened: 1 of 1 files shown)


etorreborre:
I think that this is ok because I suspect that very few people are using those modules. Thanks.

PR #1249, opened 2024-06-09
feat: introduce `FileSystem.unjarOnce`

NTPape:
`specs2-html` currently copies all of its html resources (41 files) for each specification (executed with html output), reading and traversing the `specs2-html.jar` four times in the process, overwriting its own previously copied files over and over. This is inefficient and causes unnecessary strain on the disk usage. The newly introduced `unjarOnce` method is used by `specs2-html` to only unjar its resources once for each target location and filter criteria.

I went with a simple LRU cache as to not risk OOM, but I'm open to suggestions.

(I would also backport this to v4 if / once this is merged.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.4.2]
+        scala: [3.3.3]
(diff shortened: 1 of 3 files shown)


etorreborre:
Let's close this PR, since it was included in https://github.com/etorreborre/specs2/pull/1250.

Issue #980, opened 2021-08-31
specs2-commons depends on parser-combinators 2.0.0


melezov:
Ever since 4.12.2, specs2-commons is depending on scala-parser-combinators 2.x:
https://mvnrepository.com/artifact/org.specs2/specs2-common_2.13/4.12.2

    +-org.specs2:specs2-common_2.13:4.12.3 [S]
      +-org.portable-scala:portable-scala-reflect_2.13:1.1.1 [S]
      +-org.scala-lang.modules:scala-parser-combinators_2.13:2.0.0 [S]
      +-org.scala-lang:scala-reflect:2.13.6 [S]
      +-org.scala-sbt:test-interface:1.0
      +-org.specs2:specs2-fp_2.13:4.12.3 [S]

I'm raising this in light of the recent [scala-xml revert to 1.3.x line](https://github.com/etorreborre/specs2/releases/tag/SPECS2-4.12.5) in 4.12.5 as it might be something to reconsider.
Also, the code mentions that combinators 2.x are a requirement for `scala-matcher-extra` (https://github.com/etorreborre/specs2/blob/master/project/depends.scala#L18), but as you can see above it's being used within commons, so everywhere

scala-parser-combinators 2.x seems to be binary incompatible with 1.x as posted here:
https://github.com/scala/scala-parser-combinators/releases/tag/v2.0.0

This causes issues when adding specs2 to e.g. Play projects.
Perhaps explicitly overriding the version to either 2.x or 1.x would work in this type of setup. Will test on our CI.

    [error] java.lang.RuntimeException: found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
    [error]
    [error]         * org.scala-lang.modules:scala-parser-combinators_2.13:2.0.0 (early-semver) is selected over {1.1.2, 1.1.2, 1.1.2}
    [error]             +- org.specs2:specs2-common_2.13:4.12.3               (depends on 2.0.0)
    [error]             +- org.playframework.anorm:anorm_2.13:2.6.10          (depends on 1.1.2)
    [error]             +- com.typesafe.play:play_2.13:2.8.8                  (depends on 1.1.2)
    [error]             +- com.typesafe:ssl-config-core_2.13:0.4.2            (depends on 1.1.2)

I don't have a suggestion on how to proceed, raising for awareness.

etorreborre:
[...]

PR #1315, opened 2025-03-14
scala-parser-combinators v1 for Scala 2.12, v2 for 2.13+

mkurz:
Hi @etorreborre!

Currently you depend on scala-parser-combinators v1 for _all_ Scala 2 versions. However, after looking deeper into this, I am pretty sure we just need to stay on v1 for Scala 2.12, but there is no reason to not upgrade scala-parser-combinator to v2 for Scala 2.13+. Because only Scala 2.12 itself uses scala-parser-combinators v1, however Scala 2.13 and 3 do not. I found out during looking at following issue - please see the comment I added there for further explanation:
- https://github.com/playframework/playframework/issues/13179

E.g. also look at other libraries, they do exactly the same meanwhile:
- https://github.com/eed3si9n/scalaxb/blob/v1.12.2/project/dependencies.scala#L69-L72
- https://github.com/sbt/sbt-pgp/blob/v2.3.1/project/Dependencies.scala#L17-L22
- https://github.com/sbt/sbt-native-packager/blob/v1.11.1/build.sbt#L54-L62

Please also see the comment I just added in
- https://github.com/etorreborre/specs2/issues/980#issuecomment-2725932324

I tested locally, with this PR merged I can build Play and run Play applications without version conflicts - of cours we from Play would need a new specs2 version:
- https://github.com/playframework/playframework/pull/13197

@etorreborre I am like 100% sure this is correct. I keep this as draft currently, but would you be able to merge that later and release `SPECS2-4.21.0` with it?

### DIFF
--- a/project/depends.scala
+++ b/project/depends.scala
@@ -41,7 +41,12 @@ object depends {
     ))
 
   def scalaParser = Def.setting {
-    Seq("org.scala-lang.modules" %%% "scala-parser-combinators" % "1.1.2")
+    Seq("org.scala-lang.modules" %%% "scala-parser-combinators" % {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, 12)) => "1.1.2"
+        case _             => "2.4.0"
+      }
+    })
   }
   def scalaParserNative = Def.setting {
     Seq("org.scala-lang.modules" %%% "scala-parser-combinators" % "2.4.0")


etorreborre:
Yes I am. Sorry for the delay, just busy with non-open source life at the moment :-D.

etorreborre:
I don't know why the build is now failing though.

mkurz:
It's because in your workflow you use `ubuntu-latest`. GitHub changed it from their ubuntu 20.04 (or 22.04 can't remember) image to 24.04 which does not include sbt anymore.

Give me a sec, I will fix that for your (pr incoming).

mkurz:
@etorreborre No CI anymore?

mkurz:
Seems I made a mistake int he yml "every step must define a `uses` or `run` key"

mkurz:
> Seems I made a mistake int he yml "every step must define a `uses` or `run` key"

- Fix: #1319

etorreborre:
@mkurz I published `4.21.0`.

PR #1345, opened 2025-06-29
Provide an `IsEmpty` instance for `IterableOnce`

jcazevedo:
Is it OK to have an `IsEmpty` instance for `IterableOnce`? That way we can avoid some specific instances and also allow `beEmpty` to be used on some additional types.

## Summary by CodeRabbit

* **Refactor**
  * Simplified emptiness checks by consolidating multiple collection-specific implementations into a single, more general approach for all iterable types.
* **Chores**
  * Removed special handling for options and specific collections, streamlining internal logic.


etorreborre (approved):
Great idea, thanks!

etorreborre:
This is published as `5.6.4`

jcazevedo:
> This is published as `5.6.4`

Thank you for this (and for the bincompat fixes!).

Issue #1356, opened 2025-08-04
Awaiting on `Future` via suggested extensions is not very practical with type hierarchies


bdmendes:
The following example should be illustrative:

```scala
class A
class B extends A

def foo: Future[A] = ???

Await.result(foo, 1.second) === new B // ok

Await.result(foo, 1.second) must beEqualTo(new B) // ok

foo must beEqualTo(new B).await // does not compile
                                // Found: org.specs2.matcher.EqualityMatcher[B]
                                // Required: org.specs2.matcher.Matcher[A]
```

This makes testing async APIs tedious, as one needs to:

```scala
foo must beEqualTo[A](new B).await
```

I am not sure what's the best way to fix it. I tried:

```scala
extension [F[_] <: Matcher[_], T](m: F[T])(using ee: ExecutionEnv) {
  def awaitMine: F[Future[T]] = ???
}
```

But the problem is not that the extension is not the generic type, but rather, I think, that the compiler does not know that it needs to widen `B` to `A` before attempting the extension.

etorreborre:
That's annoying indeed. I can't work on this right now but maybe something needs to be done with the `must` method as well.

bdmendes:
Yeah, I thought about providing a `Conversion` from `Matcher[T]` to `Matcher[U]` if `U >: T`, but then all strong safety in v5 would be thrown away (would be able to move up to `Any` everywhere...). That might work however if scoped to the `must` parameter, as in, if I need a `Matcher[U]` and have a `Matcher[T]`, I might derive it in the `must` implementation somehow (maybe throw and catch a failure if `U` turns out not to be a `T` at runtime?). Not really sure; I'm not familiar with all the internals here, but it would be great to see something like this.

PR #1359, opened 2025-08-17
Summon a `Matcher[U]` for `U >: T` where `Matcher[T]` is available

bdmendes:
Fixes https://github.com/etorreborre/specs2/issues/1356.
This makes sense to me, but it does mean that something like
```scala
Unrelated1(...) must beEqualTo(Unrelated2(...)).someMethod
```
will now compile and only fail at runtime. This does not seem that bad, since
```scala
Unrelated1(...) must beEqualTo(Unrelated2(...))
```
already compiles as of now.
I've tried to set some constraint against `Any` and/or `AnyRef`, but then the implicit conversion stopped being considered. Please let me know if this can be improved!

### DIFF
--- a/matcher/shared/src/main/scala/org/specs2/matcher/Matcher.scala
+++ b/matcher/shared/src/main/scala/org/specs2/matcher/Matcher.scala
@@ -285,6 +285,8 @@ trait MatcherCreation:
       def f1(t: =>T) = f(t)
       f1
 
+  /** this allows a matcher of a subtype to be used where a matcher of the supertype is expected
+    */
   given widenToSuperMatcher[T: ClassTag, U >: T]: Conversion[Matcher[T], Matcher[U]] with
     def apply(outer: Matcher[T]): Matcher[U] =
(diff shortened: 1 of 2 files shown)


bdmendes:
It seems that this does not work for all cases. Anyway, hopefully it's a place for discussion 😛

etorreborre:
I think that this is a limit of Scala's type inference. The expression for which we would like to find consistent type parameter values amounts to:
```
must(foo)(await(beEqualTo(new B))(using ee))
```
But I think that the type for the first parameter list is determined independently from the type in the second parameter list. 
The first one gives `Future[A]` and the second one `Matcher[Future[B]]` and `must` can not unify `Future[A]` with `Future[B]` because `A >: B`.

bdmendes:
Yes. This is an attempt to circumvent that behavior via explicit widening of the resulting matcher.
(I am not sure if this should be fixed in the compiler side. It makes sense to me that it works this way.)

PR #1366, opened 2025-09-02
[specs2-4.x-maintenance] Upgrade mockito-core to latest version (5.19.0)

mkurz:


mkurz:
@etorreborre Mockito requires Java 11 since 2,5 years ago already when it dropped support for Java 8 with Mockito version 5:
- https://github.com/mockito/mockito/commit/a7c7fdb4f972d7fb6736f692e5ee208034f214f3
- https://github.com/mockito/mockito/releases/tag/v5.0.0

Do you think it's time for the 4.x maintainance branch to also drop support for Java 8?
I would vote for yes.

mkurz:
Main reason is why I want to upgrade mockito because it has a dependency on an outdated byte-buddy dependency which does not support java 25 yet. Latest mockito release does use latest byte buddy however.

mkurz:
Green!

